PR #2, opened 2018-02-06
Quest bundles

supersam654:
I think it's kinda silly to have separate repos for every spec so I'm repurposing this repo as the `qwiklabs-bundle-specs` repo (we can rename it later).

The only interesting file in this change is [`quest-bundle-spec.md`](https://github.com/CloudVLab/qwiklabs-lab-bundle-spec/compare/quest-bundles?expand=1#diff-2bb44c91feb0036d6989a1f7a7b23df5).


supersam654:
[Define what a quest bundle looks like](https://trello.com/c/skg6ms4S/1634-define-what-a-quest-bundle-looks-like)

supersam654:
@jgan note/comment on:

* Resources in `whisperer.yaml` are not prefixed by anything. Git Whisperer is responsible for seeing if a content ID matches a resource and adding `type: Resource` or `type: Lab` appropriately.
  * If the resource is not present, GW automatically assumes it's a lab
* `qwiklabs.yaml` includes `type: {Resource|Lab}` so unbundling does not have to guess what a reference ID refers to.
* `whisperer.yaml` lets you omit `path` in an `options` block if all of your options are "OR"s. `path` is added in `qwiklabs.yaml` by GW

In `whisperer.yaml`
```yaml
  - choice:
      prompt:
        locales:
          en: Learn about compute options
      options:
        - content: intro-to-appengine-python
        - content: intro-to-kubernetes-engine
        - content: intro-to-cloud-functions
```

converted in `qwiklabs.yaml`

```yaml
  - choice:
      prompt:
        locales:
          en: Learn about compute options
      options:
        - path:
          - type: Lab
            content: intro-to-appengine-python
        - path:
          - type: Lab
            content: intro-to-kubernetes-engine
        - path:
          - type: Lab
            content: intro-to-cloud-functions
```

Unrelated to anything in particular in this PR, I believe that labs **should not** be bundled inside of Quest Bundles and should exclusively be pulled from Alexandria. Bloat and versioning confusion being my two biggest concerns.

PR #3, opened 2018-02-12
Quest bundles josh

jgan:


examples/quest-1-simple/qwiklabs.yaml
```diff
@@ -0,0 +1,51 @@
...
+      <p>This course is intended for people who</p>
+      <ul><li>Do not understand the power of frienship</li></ul>
+
+prerequisites:
+  locales:
+    en: <p>A firm understanding of interpersonal interactions is recommended but not required.</p>
+
+suggested_duration:
+  locales:
+    en: 1 day
```
supersam654: Did you have a particular use case in mind where different locales would take different amounts of time?

examples/quest-1-simple/whisperer.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+# Quest Attributes
+id: my-awesome-quest
+
+tags: [sample, life-changing, gcp]
+logo: ./quest-logo.png
+default_locale: en
+
+steps:
+  - content: lab-minimal
```
supersam654: Do you think this will uniquely identify any piece of content (in the bundle) or would we have to include a `type: {lab|file|url}` as well?

examples/quest-2-step-options/qwiklabs.yaml
```diff
@@ -0,0 +1,89 @@
...
+    locales:
+      en:
+        title: Wasn't that fun!
+        uri: https://youtu.be/dNW0B0HsvVs
+      es:
+        title: Eso fue divertido, ¿verdad?
+        uri: https://youtu.be/dNW0B0HsvVs
+
+steps:
+  - content: $intro-video
```
supersam654: I immediately knew this syntax was referring to a resource defined in the bundle (and I like it), but I'm concerned that special characters will raise the barrier to entry for some people.

examples/quest-2-step-options/qwiklabs.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - content: intro-to-gcp
+  - choice:
+      prompt:
+        locales:
+          en: Learn about compute options
+      options:
+        - content: intro-to-appengine-python
```
supersam654: In `quest-2-step-options/whisperer.yaml`, steps nested under `options` have a `path` key:

```
  - choice:
      options:
        - path:
          - content: intro-to-appengine-python
```

But here you've omitted the `path` key (and the choices are simpler). Do you intend for `path` to be an optional nested attribute that is only required for complicated choices or was this just an accidental omission?

examples/quest-3-path-options/whisperer.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - content: $intro-video
+
+  - content: intro-to-gcp
+
+  - choice:
+      prompt:
+        locales:
+          en: Learn more about one of these common compute/database pairs.
+      options:
+        - path:
```
supersam654: Are nested `path`s just as powerful as top-level `path`s? And if so, if there a limit to how far they can be nested?

examples/quest-1-simple/qwiklabs.yaml
jgan: I wasn't thinking about using this input for different amounts of time per locale, but to allow the author to give a much more flexible input than N units. So we don't have to peg it to a specific unit and make the author do the conversion.

- one day
- one week
- a few hours
- 2-3 days
- A semester... seriously, pace yourself.
- A lifetime

examples/quest-2-step-options/qwiklabs.yaml
jgan: The former: trying to make the syntax as reabable as possible.

I expect the qwiklabs.yaml will explicitly state a type here:

```yaml
- type: choice
  options:
    - type: path
    - steps:
      - type: Lab
        content: intro-to-appengine-python
```

examples/quest-2-step-options/qwiklabs.yaml
jgan: Honestly, I agree.

I'm looking to find a balance for making both resources and labs easy to reference without making it a mess to interpret.

examples/quest-3-path-options/whisperer.yaml
jgan: I use `steps` and `path` interchangeably. I was playing around to see which I liked better; but before we finalize this, they should all be unified.

examples/quest-1-simple/whisperer.yaml
jgan: I think it's actually identifying content in the same library as where this quest bundle is pulled from... which for the purposes of git-whisperer means a folder name in the `labs` folder.

examples/quest-2-step-options/qwiklabs.yaml
supersam654: I was philosophically against Git Whisperer rewriting yaml structure in lab bundles but there aren't lab bundles and the specs already feel like they're diverging so I'll allow it.

jgan: I *could* be convinced that qwiklabs.yaml doesn't require explicit types in the array, and we interpret it by what keys actual exist... or alternatively that we don't mind subjecting lab authors to the more verbose syntax.

As a pseudo-related aside, how do you feel about this situation?

```yaml
- type: choice
  options:
    - type: path
      steps:
        - type: Resource
          ref_id: my-appengine-intro-video
        - type: Lab
          content: intro-to-appengine-python
        - type: Lab
          content: intro-to-datastore-python
    - type: Lab
      content: massive-lab-that-is-equivalent-to-other-path
```

jgan: Your original question was did I forget to add a path node. I think that confusion is reasonable.

- Do you think the mixture demonstrated above is ok or just begging for confused authors?
- Should we mandate that all options be the same type?
- Should we mandate that all options be a path even if they contain only one item?

examples/quest-2-step-options/qwiklabs.yaml
jgan: How about this...?

```yaml
steps:
  - resource: intro-video
  - lab: intro-to-gcp
  ...
```

examples/quest-1-simple/qwiklabs.yaml
supersam654: I definitely see the benefit to that but I'm concerned about automatically gleaning information from a pseudo-free-form text field (not that we do that now with an integer duration).

examples/quest-2-step-options/qwiklabs.yaml
supersam654: How about `resources/intro-video` instead (where `resources` is now a reserved name and nobody can make a library called `resources` now).?

examples/quest-2-step-options/qwiklabs.yaml
supersam654: Adding `type` everywhere makes things much more verbose (and I already don't like how verbose my solution is). With that said, I prefer the syntax you used in `quest-3` (with nested paths)

PR #4, opened 2018-02-12
Translated Josh step syntax to Sam step syntax

supersam654:
I renamed every `whisperer.yaml` to `whisperer.josh.yaml` and then wrote my versions in `whisperer.sam.yaml`. The only thing I changed was things under `steps:`.

Some things I noticed while writing my version:

* It's noticeably more verbose than your way
* Despite commenting about my dislike of special characters, I added a `%` in some places
* I opted to not have Git Whisperer rewrite any YAML structure so many `next` keys have an array of one element. If we decide that rewriting structure is okay, I'd remove the array part.
* GW will require significant validations to ensure people haven't really screwed up step paths (in particular, infinite loops and dangling steps that don't lead to a `done` step).


supersam654:
Closing because I lost the bake off :(

PR #6, opened 2018-11-08
Add quiz spec to examples/ directory

rickcnagy:
This is what we have so far for definining quiz versions. We likely also
want to do something to define quizzes (i.e. qwiklabs.yaml in
examples/quiz) but we haven't figured that out yet.

### DIFF
--- a/examples/quiz/en.yaml
+++ b/examples/quiz/en.yaml
@@ -1,0 +1,43 @@
+schema_version: 1
+
+locale: en
+passing_percentage: 80
+
+items:
+  - type: "multiple-choice"
+    stem: "Who won the Nobel Prize for discovering the cause of the photoelectric effect?"
+    options:
+      - title: "Albert Einstein"
+        is_answer: true
+        rationale: "Correct! Albert Einstein won the Nobel Prize in 1922 for work that forms the foundation for how modern solar panels work."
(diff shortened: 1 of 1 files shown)


examples/quiz/en.yaml
```diff
@@ -0,0 +1,43 @@
+schema_version: 1
+
+locale: en
```
jgan: Having one locale per .yml file is cool for the upload interface and the GitWhisperer implementation, but it's contrary to the content bundle design specs.

Check out the example lab's use of default locale:

https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/3cdcd1909acff93a1617836273e89612dfcb40bd/examples/lab-robust/qwiklabs.yaml#L5

... and localizable properties:

https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/3cdcd1909acff93a1617836273e89612dfcb40bd/examples/lab-robust/qwiklabs.yaml#L7-L15

jgan: @rickcnagy - poke

rickcnagy:
@jgan - I’ve been thinking about this, let’s discuss tomorrow...

rickcnagy:
Closing in favor of #15 .

PR #7, opened 2018-11-28
Factor out a resource spec

joncass:
Overall goal: preliminary cleanup before solidifying Course spec

Pieces:
- Factor out `resources` into its own spec
- Add the word "course" in, in various places
- Misc cleanup to make all of the examples and documentation match each other and reality


joncass:
@jgan @steve-rivera @davetchen @rickcnagy FYI

PR #15, opened 2018-12-10
Add quiz bundle spec with examples

rickcnagy:
This includes the full quiz spec with a robust and minimal example.

### DIFF
--- a/examples/quiz-robust/qwiklabs.yml
+++ b/examples/quiz-robust/qwiklabs.yml
@@ -1,0 +1,105 @@
+entity_type: Quiz
+schema_version: 1
+
+title: Robust Quiz
+passing_percentage: 67
+
+items:
+  - type: multiple-choice
+    stem:
+      locales:
+        en: Who won the Nobel Prize for discovering the cause of the photoelectric effect?
+        fr: Qui a remporté le prix Nobel pour avoir découvert la cause de l'effet photoélectrique?
+    options:
+      - title:
+          locales:
+            en: Albert Einstein
+            fr: Albert Einstein
+        is_answer: true
+        rationale:
+          locales:
+            en: Correct! Albert Einstein won the Nobel Prize in 1922 for work that forms the foundation for how modern solar panels work.
+            fr: Correct! Albert Einstein a reçu le prix Nobel en 1922 pour des travaux qui constituent le fondement du fonctionnement des panneaux solaires modernes.
+      - title:
+          locales:
+            en: Francis Crick
+            fr: Francis Crick
+        is_answer: false
+        rationale:
+          locales:
+            en: While Crick did win a Nobel Prize, it was for his discovery of DNA along with James Watson.
+            fr: Bien que Crick ait remporté un prix Nobel, c'était pour sa découverte de l'ADN avec James Watson.
+      - title:
+          locales:
+            en: Rosalind Franklin
+            fr: Rosalind Franklin
+        is_answer: false
+        rationale:
+          locales:
+            en: Sorry! While her work was instrumental in discovering DNA, Rosalind Franklin never won a Nobel Prize.
+            fr: Désolé! Alors que son travail a joué un rôle déterminant dans la découverte de l'ADN, Rosalind Franklin n'a jamais remporté de prix Nobel.
+      - title:
+          locales:
+            en: Max Planck
+            fr: Max Planck
+        is_answer: false
+        rationale:
+          locales:
+            en: Max Planck won the Nobel Price in Physics in 1918 for his work on quantum theory, and while his work was used in the discovery of the photoelectric effect, he actually rejected the initial theory.
+            fr: Max Planck a remporté le prix Nobel de physique en 1918 pour ses travaux sur la théorie quantique. Bien que ses travaux aient servi à la découverte de l'effet photoélectrique, il a en réalité rejeté la théorie initiale.
+  - type: multiple-select
+    stem:
+      locales:
+        en: Which of the following is a state in the United States?
+        fr: Lequel des États suivants est un État aux États-Unis?
+    options:
+      - title:
+          locales:
+            en: Massachusetts
+            fr: Massachusetts
+        is_answer: true
+        rationale:
+          locales:
+            en: Yep! Massachusetts is located in the Northeast.
+            fr: Oui! Le Massachusetts est situé dans le nord-est.
+      - title:
+          locales:
+            en: Manitoba
+            fr: Manitoba
+        is_answer: false
+        rationale:
+          locales:
+            en: Sorry! Manitoba is a Canadian province, though it does share a border with the Midwest United States.
+            fr: Désolé! Le Manitoba est une province canadienne, même s'il partage une frontière avec le Midwest américain.
+      - title:
+          locales:
+            en: Oregon
+            fr: l'Oregon
+        is_answer: true
+        rationale:
+          locales:
+            en: Yep! Oregon is a state in the Pacific Northwest.
+            fr: Oui! L'Oregon est un état du nord-ouest du Pacifique.
+      - title:
+          locales:
+            en: Washington, D.C.
+            fr: Washington, D.C.
+        is_answer: false
+        rationale:
+          locales:
+            en: Washington, D.C. is specifically mentioned in the U.S. Constitution as as a district administered by the federal government for the purpose of being the Seat of the Government of the United States. It lacks several of the rights states have, such as full representation in Congress.
+            fr: Washington, DC, est expressément mentionné dans la Constitution américaine comme un district administré par le gouvernement fédéral dans le but d’être le siège du gouvernement des États-Unis. Il manque plusieurs des droits des États, tels que la pleine représentation au Congrès.
+  - type: true-false
+    stem:
+      locales:
+        en: Direct democracy is a form of government where a single leader has ultimate ruling authority.
+        fr: La démocratie directe est une forme de gouvernement où un seul chef a l'autorité suprême.
+    answer: false
+    true_rationale:
+      locales:
+        en: Sorry! A single leader having ultimate authority describes an autocracy.
+        fr: Désolé! Un seul chef ayant l'autorité ultime décrit une autocratie.
+    false_rationale:
+      locales:
+        en: Right! That form of government would be closer to an autocracy, not democracy.
+        fr: C'est ça! Cette forme de gouvernement serait plus proche d'une autocratie, pas d'une démocratie.
--- a/examples/quiz-robust/qwiklabs.yml
+++ b/examples/quiz-robust/qwiklabs.yml
@@ -1,10 +1,9 @@
 entity_type: Quiz
-schema_version:
+schema_version: 1
 
-default_locale: en
-
-title: Robust Quiz
+id: quiz-robust
 passing_percentage: 67
+default_locale: en
 
 items:
   - type: multiple-choice
--- a/examples/quiz-robust/qwiklabs.yml
+++ b/examples/quiz-robust/qwiklabs.yml
@@ -1,5 +1,7 @@
 entity_type: Quiz
-schema_version: 1
+schema_version:
+
+default_locale: en
 
 title: Robust Quiz
 passing_percentage: 67
--- a/quiz-bundle-spec.md
+++ b/quiz-bundle-spec.md
@@ -25,7 +25,10 @@ items: ...
 Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
 
 ### Items
-Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute.
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute. The allowed values for `type` are:
+- `multiple-choice`
+- `multiple-select`
+- `true-false`
 
 #### Multiple Choice Items
 
@@ -34,7 +37,7 @@ A quiz item that has multiple options and one answer. One option is the _answer_
 attribute               | required | type              | notes
 ----------------------- | -------- | ----------------- | -----------------------------------------
 id                      | ✓        | string            | A unique identifier for this item
-type                    | ✓        | 'multiple-choice' | The item type, which is always 'multiple-choice'
+type                    | ✓        | string            | The item type, which is always 'multiple-choice'
 stem                    | ✓        | dictionary        | A locale dictionary of the text that asks the question, such as "Which of the following is a color?"
 options                 | ✓        | array             | An array of item options
 -- id                   | ✓        | string            | A unique ID for this item option
@@ -57,7 +60,7 @@ options                 | ✓        | array             | An array of item opti
 -- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
 -- is_answer            | ✓        | string            | Whether this option is an answer or distractor
 
-### True False Items
+#### True False Items
 
 A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractor.
 
--- a/quiz-bundle-spec.md
+++ b/quiz-bundle-spec.md
@@ -1,0 +1,71 @@
+# Qwiklabs Quiz Bundle Specification
+
+**Version 1**
+
+> This is a DRAFT document. We welcome feedback as this format evolves.
+
+## `qwiklabs.yaml` Structure
+
+Here's a sample `qwiklabs.yaml` file with all nested details removed to make it easier to see the general file structure.
+
+```yml
+entity_type: Quiz
+schema_version: 1
+
+# The authoring title of the quiz - not visible to learners.
+title: Super Hard Quiz
+
+# The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
+passing_percentage: 67
+
+# The "items" (i.e. questions) contained in this quiz.
+items: ...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
+### Items
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute.
+
+#### Multiple Choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-choice' | The item type, which is always 'multiple-choice'
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | boolean           | Whether this option is an answer or distractor
+
+#### Multiple Select Items
+
+A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-select' | The item type, which is always 'multiple-select'
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | string            | Whether this option is an answer or distractor
+
+### True False Items
+
+A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractors.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'true-false'      | The item type, which is always 'true-false'
+stem                    | ✓        | string            | The text that asks the question, such as "True or false: the world is round."
+answer                  | ✓        | boolean           | The correct answer. true for "True", false for "False"
+true_rationale          | ✓        | string            | Explains why "True" is correct or incorrect
+false_rationale         | ✓        | string            | Explains why "False" is correct or incorrect
--- a/quiz-bundle-spec.md
+++ b/quiz-bundle-spec.md
@@ -12,14 +12,10 @@ Here's a sample `qwiklabs.yaml` file with all nested details removed to make it 
 entity_type: Quiz
 schema_version: 1
 
+id: final-exam
+passing_percentage: 67
 default_locale: en
 
-title: Super Hard Quiz
-
-# The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
-passing_percentage: 67
-
-# The "items" (i.e. questions) contained in this quiz.
 items: ...
 ```
 
@@ -29,13 +25,11 @@ Note that all of the localized content (stem, option titles, rationales, etc) ar
 
 attribute          | required | type    | notes
 -------------------| -------- | ------- | -----------------------------------------
-title              | ✓        | string  | The authoring title of the quiz - not visible to learners.
+default_locale     | ✓        | string  | Corresponds to the locale that the quiz is authored in. Authoring tools can use this as a hint to notify localizers when content in the default locale is updated. Also, it provides a hint to the learner interface about which locale to display if an instruction/resource is not localized for the learner's current locale.
+schema
+id                 | ✓        | string  | A unique identifier for this quiz, must be unique per "library" and URL friendly (think github org/repo)
 passing_percentage | ✓        | integer | The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
 items              | ✓        | array   | An ordered array of `items` (see below for details) in this quiz - items will appear to students in this order
-
-### Default locale
-
-The quiz bundle must specify a `default_locale`. It corresponds to the locale that the quiz is authored in. Authoring tools can use this as a hint to notify localizers when content in the default locale is updated. Also, it provides a hint to the learner interface about which locale to display if an instruction/resource is not localized for the learner's current locale.
 
 ### Items
 Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of dictionaries with appropriate attributes for the given `type`. The allowed values for `type` are:
--- a/quiz-bundle-spec.md
+++ b/quiz-bundle-spec.md
@@ -26,7 +26,7 @@ Note that all of the localized content (stem, option titles, rationales, etc) ar
 attribute          | required | type    | notes
 -------------------| -------- | ------- | -----------------------------------------
 default_locale     | ✓        | string  | Corresponds to the locale that the quiz is authored in. Authoring tools can use this as a hint to notify localizers when content in the default locale is updated. Also, it provides a hint to the learner interface about which locale to display if an instruction/resource is not localized for the learner's current locale.
-schema
+schema_version     | ✓        | integer   | Which version of the quiz bundle schema you are using
 id                 | ✓        | string  | A unique identifier for this quiz, must be unique per "library" and URL friendly (think github org/repo)
 passing_percentage | ✓        | integer | The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
 items              | ✓        | array   | An ordered array of `items` (see below for details) in this quiz - items will appear to students in this order
--- a/quiz-bundle-spec.md
+++ b/quiz-bundle-spec.md
@@ -29,22 +29,22 @@ Items are polymorphic - i.e. there are several different _item types_ that are d
 
 #### Multiple Choice Items
 
-A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_.
 
 attribute               | required | type              | notes
 ----------------------- | -------- | ----------------- | -----------------------------------------
 id                      | ✓        | string            | A unique identifier for this item
 type                    | ✓        | 'multiple-choice' | The item type, which is always 'multiple-choice'
-stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+stem                    | ✓        | dictionary        | A locale dictionary of the text that asks the question, such as "Which of the following is a color?"
 options                 | ✓        | array             | An array of item options
 -- id                   | ✓        | string            | A unique ID for this item option
 -- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
 -- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
--- is_answer            | ✓        | boolean           | Whether this option is an answer or distractor
+-- is_answer            | ✓        | boolean           | `true` if this option is an answer, and `false` if it is a distractor
 
 #### Multiple Select Items
 
-A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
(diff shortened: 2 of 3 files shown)


quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+passing_percentage: 67
+
+# The "items" (i.e. questions) contained in this quiz.
+items: ...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
+### Items
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute.
```
joncass: Please list the allowed `type` attributes here.
```suggestion
Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute. The allowed values for `type` are `multiple-choice`, `multiple-select`, and `true-false`.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+#### Multiple Choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-choice' | The item type, which is always 'multiple-choice'
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
```
joncass: You've written "A locale dictionary of locales for..." a lot of places. I think we can assume that that's what a locale dictionary is, and just write something like
```suggestion
stem                    | ✓        | dictionary        | A locale dictionary of the text that asks the question, such as "Which of the following is a color?"
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+options                 | ✓        | array             | An array of item options
```
joncass: Consider moving the `id`, `title`, `rationale`, `is_answer` to an "option spec" section below and removing it from here and the `multiple-select`

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | boolean           | Whether this option is an answer or distractor
```
joncass:
```suggestion
-- is_answer            | ✓        | boolean           | `true` if this option is an answer, and `false` if it is a distractor
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
+### Items
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute.
+
+#### Multiple Choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
```
joncass: I would save the commentary on rationale for the "option spec" that I propose below.
```suggestion
A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | boolean           | Whether this option is an answer or distractor
+
+#### Multiple Select Items
+
+A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
```
joncass:
```suggestion
A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_.
```
see above

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-select' | The item type, which is always 'multiple-select'
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
```
joncass: See above; consider factoring this out into an "option spec"

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+### True False Items
+
+A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractors.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'true-false'      | The item type, which is always 'true-false'
+stem                    | ✓        | string            | The text that asks the question, such as "True or false: the world is round."
```
joncass: I think this should be a locale dictionary
```suggestion
stem                    | ✓        | dictionary        | A locale dictionary of text that asks the question, such as "True or false: the world is round."
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+answer                  | ✓        | boolean           | The correct answer. true for "True", false for "False"
```
joncass: I think this is clearer with "code" annotation
```suggestion
answer                  | ✓        | boolean           | The correct answer. `true` for "True", `false` for "False"
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+## `qwiklabs.yaml` Structure
+
+Here's a sample `qwiklabs.yaml` file with all nested details removed to make it easier to see the general file structure.
+
+```yml
+entity_type: Quiz
+schema_version: 1
+
+# The authoring title of the quiz - not visible to learners.
+title: Super Hard Quiz
```
supersam654: I think the fact that learners will never see this title is fairly significant and I'd prefer a more descriptive key like `internal_title` that clearly indicates that this is for internal bookkeeping purposes only.

supersam654: After thinking about my comment a bit more, I believe it will be fairly controversial. I would not strongly advocate to make the change and I realize that the change would cause the quiz spec to deviate from our other specs. However, I think it's an important point and would appreciate a bit of discussion around it.

examples/quiz-robust/qwiklabs.yml
```diff
@@ -0,0 +1,105 @@
...
+    stem:
+      locales:
+        en: Who won the Nobel Prize for discovering the cause of the photoelectric effect?
+        fr: Qui a remporté le prix Nobel pour avoir découvert la cause de l'effet photoélectrique?
+    options:
+      - title:
+          locales:
+            en: Albert Einstein
+            fr: Albert Einstein
+        is_answer: true
```
supersam654: Prefixing these with `is_` feels out of place in the spec. I'd lean towards just `answer: true` and `answer: false`.

Depending on where quizzes are going and how complicated questions are going to get, you may want to change this to something like `answer_type: {answer,distractor,hard_distractor}`

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
+# Qwiklabs Quiz Bundle Specification
+
+**Version 1**
+
+> This is a DRAFT document. We welcome feedback as this format evolves.
```
supersam654: I know you must have copied this from another file, but it's never too late to bring up weirdness.

When this gets merged to master, is this file really a draft? Sure, there may be some tweaks as you write the actual implementation, but this is going to be sorta-kinda codified as V1 when the website implementation unbundles these in Dec/Jan.

quiz-bundle-spec.md
supersam654: As someone who is very far from a domain expert on test creation, I find the word `stem` to be jarring and unintuitive. I realize this is a machine-readable spec that favors verbosity and precision over convention, but I'd still appreciate a more lay word like `question` (especially because this word is almost definitely going to get propagated to the GW version).

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute.
+
+#### Multiple Choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. Each option has a required rationale that explains why that option is correct or incorrect.
+
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-choice' | The item type, which is always 'multiple-choice'
```
supersam654: Out of curiosity, why isn't the format for specifying MC questions a superset of the format for specifying T/F questions?  It feels overly complex that T/F questions don't look like MC questions with a different flag or something.

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | string            | Whether this option is an answer or distractor
+
+### True False Items
+
+A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractors.
```
supersam654:
```suggestion
A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractor.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+attribute               | required | type              | notes
+----------------------- | -------- | ----------------- | -----------------------------------------
+id                      | ✓        | string            | A unique identifier for this item
+type                    | ✓        | 'multiple-select' | The item type, which is always 'multiple-select'
+stem                    | ✓        | dictionary        | A locale dictionary of locales for the text that asks the question, such as "Which of the following is a color?"
+options                 | ✓        | array             | An array of item options
+-- id                   | ✓        | string            | A unique ID for this item option
+-- title                | ✓        | dictionary        | A locale dictionary of locales for the content of this option, such as "Blue"
+-- rationale            | ✓        | dictionary        | A locale dictionary of locales for an explanation of why this option is correct or incorrect
+-- is_answer            | ✓        | string            | Whether this option is an answer or distractor
```
supersam654: I'd like to see an explicit note on whether or not multiple correct answers are allowed (or if we deem that to be a spec violation).

I don't know what we technically support, but I'd be in favor of requiring exactly one correct answer (and not 0 or N). We can always relax to spec in the future to allow for multiple correct answers. I'd probably designate those probes as a different type so people don't accidentally pick multiple correct answers on a regular MC question.

quiz-bundle-spec.md
supersam654: I would not assume everyone reading the spec intuitively understands that the options are ordered just because they're in an array.

```suggestion
options                 | ✓        | array             | An ordered list of item options
```

examples/quiz-robust/qwiklabs.yml
rickcnagy: I'm not wild about `answer: true`, since that's also how we define the boolean answer to a true false item. I'm not sure I follow on `answer_type`, but I can say that if the answers got more complex it almost certainly would be an entirely new item type (which is likely to happen and is a good thing!) and not adding complexity to MC, just like how MS is not a property on MC but an entirely separate type.

quiz-bundle-spec.md
rickcnagy: I personally think we should remove these versions and draft messages on all files ASAP. We're already writing unbundling code for game templates, so these are definitely real, not draft, specs.

In the meantime I'm happy to remove it here - or just wait and do all specs at once in a follow up PR for consistency.

quiz-bundle-spec.md
rickcnagy: Hmm... that's actually a really good idea. I'm good with `internal_title` - @jgan or @joncass - any opinions?

quiz-bundle-spec.md
rickcnagy: Good question! In my experience, even though these item types all _feel_ similar, they're all actually different enough that trying to share too much code / definitions / etc just leads to _lots_ of conditional logic and validation. If we instead treat them as different but maybe share rendering mixins, then things are a bit more verbose but things are a lot clearer.

quiz-bundle-spec.md
rickcnagy: Yeah, I can see that - but actually it's used really consistenly as a term of art in the authoring world (notice it shows up 10 times on the Wiki page for MC: https://en.wikipedia.org/wiki/Multiple_choice). Happy to explain a bit more what a stem is here.

quiz-bundle-spec.md
rickcnagy: I thought about that - I thought it might be a bit confusing to authors. Then again, this is more machine generated / readable. I'll switch to that.

rickcnagy: Actually, options _aren't_ ordered, so I'll clarify that.

quiz-bundle-spec.md
rickcnagy: Yep - each MC item should have exactly 1 answer, and each MS item should have at least one answer. Will add a note for that (even if we don't validate immediately).

examples/quiz-robust/qwiklabs.yml
jgan: This may require a bit of mental gymnastics, however...

This actually matches the `true-false` use of `answer` pretty well. For each of these individual options, they each have a correct `true`/`false` value.

quiz-bundle-spec.md
jgan: Just because there exists an implementation of a spec, does not mean the spec if finalized. Even after Lab bundles were supported by GitWhispere/Website, they were still drafts. We have not guaranteed that any of the features described in this document will be supported by external systems.

Heck, the bundling/unbundling code we implemented did *not* match the Lab bundle spec described in this repo until very recently.

Now that....

- the spec has been updated to reflect reality of our implementation, and
- we've got multiple engineers contributing to this and the unbundling code has gone through a couple revisions...

... I'm willing to remove the V1 draft status at the end of this quarter.

jgan: @supersam654 and @rickcnagy  ^^

quiz-bundle-spec.md
jgan: Yeah I like the idea, but `internal_title` doesn't feel descriptive enough... and I just spent 20 minutes on thesaurus.com trying to come up with something better to no avail.

quiz-bundle-spec.md
jgan: I'm all for this suggestion. You should feel free to assume that the readers of this document are software engineers or at least Lab Architects. Content authors will read the corresponding GitWhisperer (or other integrating system's) documentation.

Along those lines feel free to get really specific about type definition...

```suggestion
options                 | ✓        | Array of Options  | Unordered list of options for the question. See [mulitple choice options spec](#multiple-choice-options).
```

quiz-bundle-spec.md
jgan: I think having 'multiple-choice' in the type column is confusing. You specify in the notes that the value must be 'multiple-choice'.

```suggestion
type                    | ✓        | string            | The item type, which is always 'multiple-choice'
```

quiz-bundle-spec.md
jgan: ... as a list please.

```suggestion
Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute. The allowed values for `type` are: 
- `multiple-choice`,
- `multiple-select`, and
- `true-false`.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+### True False Items
```
jgan: This header is off by one from it's siblings.

```suggestion
#### True False Items
```

quiz-bundle-spec.md
jgan: For not wanting to confuse non-technical authors, the word "polymorphic" is a choice bit of CS jargon. :)

Check out the [Environment Resources](https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/master/lab-bundle-spec.md#environment-resources) section of the lab bundle spec. It's far from perfect (reading it now is inspiring me to make some significant edits), but personally I find the general layout more clear.

- Why is this list of objects important?
- What do all the objects in this list have in common?
- What are the different variations of objects?

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+Here's a sample `qwiklabs.yaml` file with all nested details removed to make it easier to see the general file structure.
+
+```yml
+entity_type: Quiz
+schema_version: 1
+
+# The authoring title of the quiz - not visible to learners.
+title: Super Hard Quiz
+
```
jgan: Needs a `default_locale`. See https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/master/lab-bundle-spec.md#default-locale

No need to explain it more deeply in this PR. Let's note it as a TODO to move the default_locale description in lab bundle to a centralized localization spec file.

quiz-bundle-spec.md
```diff
@@ -0,0 +1,71 @@
...
+# The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
+passing_percentage: 67
+
+# The "items" (i.e. questions) contained in this quiz.
+items: ...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
```
jgan: Add a full specification of the top level attributes, a la [course bundle spec](https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/master/course-template-bundle-spec.md#attribute-specification)

quiz-bundle-spec.md
joncass: If you're going to do it as a bulleted list, I don't think it should have commas or "and" or a period:
```suggestion
Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute. The allowed values for `type` are:
- `multiple-choice`
- `multiple-select`
- `true-false`
```

quiz-bundle-spec.md
joncass: It also took me a bit to get used to the quiz terminology, but I think where there are well-established terms of art, we should learn them and use them (and have our authors learn and use them as well)

examples/quiz-robust/qwiklabs.yml
rickcnagy: @jgan yep, that's definitely true. I don't think we should go down that route, though - more explicit is better when it comes to these item types.

quiz-bundle-spec.md
rickcnagy: @jgan EOQ1 sounds great 😄

quiz-bundle-spec.md
rickcnagy: @jgan @supersam654 do we want to settle on `title` or `internal_title`? I'm unopinionated.

quiz-bundle-spec.md
rickcnagy: I like @joncass 's the best  🌹

rickcnagy: @jgan isn't the main audience of this spec engineers who are building to the spec, since authors will normally author via GW?

rickcnagy:
Alright @supersam654 @joncass @jgan I believe I've addressed all of your current comments - let me know how it looks now. Thanks!

examples/quiz-robust/qwiklabs.yml
```diff
@@ -0,0 +1,107 @@
+entity_type: Quiz
+schema_version:
```
joncass: missing `schema_version` value

examples/quiz-robust/qwiklabs.yml
```diff
@@ -0,0 +1,107 @@
...
+    options:
+      - title:
+          locales:
+            en: Albert Einstein
+            fr: Albert Einstein
+        is_answer: true
+        rationale:
+          locales:
+            en: Correct! Albert Einstein won the Nobel Prize in 1922 for work that forms the foundation for how modern solar panels work.
+            fr: Correct! Albert Einstein a reçu le prix Nobel en 1922 pour des travaux qui constituent le fondement du fonctionnement des panneaux solaires modernes.
```
joncass: I don't think `Correct!` is french

quiz-bundle-spec.md
joncass: I lean towards `title` but will defer if someone feels strongly

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+### Quiz attributes
+
+attribute          | required | type    | notes
+-------------------| -------- | ------- | -----------------------------------------
+title              | ✓        | string  | The authoring title of the quiz - not visible to learners.
+passing_percentage | ✓        | integer | The threshold grade that a student needs to achieve in order to count as "passing" the quiz.
+items              | ✓        | array   | An ordered array of Items in this quiz - items will appear in this order in the quiz UI
```
joncass:
```suggestion
items              | ✓        | array   | An ordered array of `items` (see below for details) in this quiz - items will appear to students in this order
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Default locale
+
+The quiz bundle must specify a `default_locale`. It corresponds to the locale that the quiz is authored in. Authoring tools can use this as a hint to notify localizers when content in the default locale is updated. Also, it provides a hint to the learner interface about which locale to display if an instruction/resource is not localized for the learner's current locale.
+
+### Items
+Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of items defined correctly for their type as defined by the `type` attribute. The allowed values for `type` are:
```
joncass:
```suggestion
Items are polymorphic - i.e. there are several different _item types_ that are defined slightly differently. `items` is an array of dictionaries with appropriate attributes for the given `type`. The allowed values for `type` are:
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+#### multiple-choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. There must be exactly one answer.
+
+attribute | required | type       | notes
+----------| -------- | -----------| -----------------------------------------
+id        | ✓        | string     | A unique identifier for this item
+type      | ✓        | string     | The item type, which is always 'multiple-choice'
+stem      | ✓        | dictionary | A locale dictionary of the text that asks the question, such as "Which of the following is a color?"
+options   | ✓        | array      | An array of Options; order does not matter
```
joncass:
```suggestion
options   | ✓        | array      | An array of `options` (see below for details); order does not matter
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### multiple-select Items
+
+A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_.
```
joncass:
```suggestion
A quiz item that has multiple options, any number of which are answers. Correct options are _answers_, and incorrect options are _distractors_.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+attribute | required | type       | notes
+----------| -------- | -----------| -----------------------------------------
+id        | ✓        | string     | A unique identifier for this item
+type      | ✓        | string     | The item type, which is always 'multiple-select'
+stem      | ✓        | dictionary | A locale dictionary for the text that asks the question, such as "Which of the following is a color?"
+options   | ✓        | array      | An array of Options; order does not matter
```
joncass:
```suggestion
options   | ✓        | array      | An array of `options` (see below for details); order does not matter
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### true-false Items
+
+A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractor. There must be at least one answer.
```
joncass:
```suggestion
A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractor.
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+- `true-false`
+
+#### multiple-choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. There must be exactly one answer.
+
+attribute | required | type       | notes
+----------| -------- | -----------| -----------------------------------------
+id        | ✓        | string     | A unique identifier for this item
+type      | ✓        | string     | The item type, which is always 'multiple-choice'
```
joncass:
```suggestion
type      | ✓        | string     | The item type, which is always `multiple-choice`
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+options   | ✓        | array      | An array of Options; order does not matter
+
+#### multiple-select Items
+
+A quiz item that has multiple options and multiple answers. Correct options are _answers_, and incorrect options are _distractors_.
+
+attribute | required | type       | notes
+----------| -------- | -----------| -----------------------------------------
+id        | ✓        | string     | A unique identifier for this item
+type      | ✓        | string     | The item type, which is always 'multiple-select'
```
joncass:
```suggestion
type      | ✓        | string     | The item type, which is always `multiple-select`
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### true-false Items
+
+A quiz item that has two predefined options: true and false. The correct option (e.g. "True") is the answer and the other is the distractor. There must be at least one answer.
+
+attribute       | required | type       | notes
+----------------| -------- | -----------| -----------------------------------------
+id              | ✓        | string     | A unique identifier for this item
+type            | ✓        | string     | The item type, which is always 'true-false'
```
joncass:
```suggestion
type            | ✓        | string     | The item type, which is always `true-false`
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+stem            | ✓        | dictionary  | A locale dictionary of text that asks the question, such as "True or false: the world is round."
+answer          | ✓        | boolean     | The correct answer. `true` for "True", `false` for "False"
+true_rationale  | ✓        | string      | Explains why "True" is correct or incorrect
+false_rationale | ✓        | string      | Explains why "False" is correct or incorrect
+
+### Option
+
+multiple-choice and multiple-select items both have an `options` array which contains all of the answer(s) and distractors that the user may choose from. Options are defined for both item types below:
```
joncass:
```suggestion
`multiple-choice` and `multiple-select` items both have an `options` array which contains all of the answer(s) and distractors that the user may choose from. Options are defined for both item types below:
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+attribute    | required | type       | notes
+-------------| -------- | -----------| -----------------------------------------
+-- id        | ✓        | string     | A unique ID for this Option
```
joncass:
```suggestion
id        | ✓        | string     | A unique ID for this Option
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+-- title     | ✓        | dictionary | A locale dictionary for the content of this option, such as "Blue"
```
joncass:
```suggestion
title     | ✓        | dictionary | A locale dictionary for the content of this option, such as "Blue"
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+-- rationale | ✓        | dictionary | A locale dictionary for an explanation of why this option is correct or incorrect
```
joncass:
```suggestion
rationale | ✓        | dictionary | A locale dictionary for an explanation of why this option is correct or incorrect
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+-- is_answer | ✓        | boolean    | `true` if this option is an answer, and `false` if it is a distractor
```
joncass:
```suggestion
is_answer | ✓        | boolean    | `true` if this option is an answer, and `false` if it is a distractor
```

examples/quiz-robust/qwiklabs.yml
rickcnagy: Well that’s not very robust.

examples/quiz-robust/qwiklabs.yml
rickcnagy: Actually, it is! It’s the masculine version of the word which according to Google Translate (and my shaky memory from HS french) is the correct (heh) gender in this context (ie where there’s no noun to modify).

quiz-bundle-spec.md
rickcnagy: In keeping with the game spec, how about authoring_title?

jgan: We could also eschew `title` and just go with something like `short_description`

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+items: ...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
+### Quiz attributes
+
+attribute          | required | type    | notes
+-------------------| -------- | ------- | -----------------------------------------
+title              | ✓        | string  | The authoring title of the quiz - not visible to learners.
```
jgan:
```suggestion
title              | ✓        | string  | The authoring title of the quiz - not visible to learners.
default_locae | ✓        | string  | see [note below](#default-locale)
```

quiz-bundle-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### multiple-choice Items
+
+A quiz item that has multiple options and one answer. One option is the _answer_, and the rest are _distractors_. There must be exactly one answer.
+
+attribute | required | type       | notes
+----------| -------- | -----------| -----------------------------------------
+id        | ✓        | string     | A unique identifier for this item
+type      | ✓        | string     | The item type, which is always `multiple-choice`
+stem      | ✓        | dictionary | A locale dictionary of the text that asks the question, such as "Which of the following is a color?"
```
jgan: @rickcnagy @joncass - I would prefer this be a "stronger" type than just dictionary. We aren't accepting an arbitrary dictionary here. It's specifically a "LocaleDictionary" (not pushing this as the official name) which has a very specific structure.

It's the same reasoning for why Options was pulled out to a separate definition, and even before that you felt compelled to put the accepted Options properties in "nested" rows.

If you want to just leave it as `dictionary` for now, and we can update it all the references across all the documents after overhauling (and centralizing) the localization documentation, that's cool with me. However, I want to call out that generic `dictionary` and `array` should be avoided throughout this repo. We don't have any instance of "unspecified" / "blobby" input yet, and I suspect we never will.

quiz-bundle-spec.md
jgan: Correct. I was just poking fun at the fact that before that had been discussed, you were already using the word "polymorphic" :stuck_out_tongue_winking_eye:

jgan:
@rickcnagy - I had a few more comments... that it's honestly hard to see in the comment stream at this point. The only critical one was adding `default_locale` to the table of Quiz properties. :+1: on the section that describes it. 

LGTM

quiz-bundle-spec.md
rickcnagy: I've replaced `title` with `id`, so this is now moot 😄

quiz-bundle-spec.md
rickcnagy: Done

quiz-bundle-spec.md
rickcnagy: ^^ I say we go for a unified "LocaleDictionary" type that's consistent across all of the specs (and then can share unbundling code as well!)

quiz-bundle-spec.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+Note that all of the localized content (stem, option titles, rationales, etc) are HTML content that may be displayed in various contexts. All of these chunks will be sanitized according to the restricted set in the [HTML spec](./html-spec.md).
+
+### Quiz attributes
+
+attribute          | required | type    | notes
+-------------------| -------- | ------- | -----------------------------------------
+default_locale     | ✓        | string  | Corresponds to the locale that the quiz is authored in. Authoring tools can use this as a hint to notify localizers when content in the default locale is updated. Also, it provides a hint to the learner interface about which locale to display if an instruction/resource is not localized for the learner's current locale.
+schema
```
joncass:
```suggestion
schema_version     | ✓        | integer   | Which version of the quiz bundle schema you are using
```

rickcnagy:
Alright all - I'm calling this done! If you have follow up issues, please open a new PR. Thanks!

PR #16, opened 2018-12-14
Specify how to reference activities

joncass:
Specify how to reference activities


course-template-bundle-spec.md
```diff
@@ -9,6 +9,7 @@
 Here's a sample `qwiklabs.yaml` file with all nested details removed to make it easier to see the general file structure.
 
 ```yml
+id: my-library/gcp-intro-course
```
jgan: :grimacing:..

Are we adding this temporarily because it reflects current reality, and therefore will remove it when we change the unbundle endpoint parameters?

joncass: yes indeed

jgan: ughh... Fine.

PR #17, opened 2018-12-19
Remove id from bundles

joncass:
Remove id from bundles


joncass:
FYI @rickcnagy @supersam654

joncass:
[Make content_id an API parameter instead of part of the bundle metadata](https://trello.com/c/BXNm1Ktj/2670-make-contentid-an-api-parameter-instead-of-part-of-the-bundle-metadata)

PR #21, opened 2019-01-04
Video duration

joncass:
- Separate Video attributes
- Add duration
- Make "example" a section


resource-spec.md
```diff
@@ -57,6 +66,7 @@ resources:
         locales:
           en: https://youtu.be/oHg5SJYRHA0
           es: https://youtu.be/7jjoyy7_RCk
+    duration: 120
```
davetchen: `course-video` looks good. `lab-video` is still in the old format

joncass:
As discussed, that is intentional so that lab videos can behave as they always have

PR #23, opened 2019-02-11
Remove id fields from quiz bundle spec

rickcnagy:
I'd previously added a top-level ID for quizzes as well as ID's on item,
item options, etc. None of that should be necessary.

### DIFF
--- a/examples/quiz-minimal/qwiklabs.yml
+++ b/examples/quiz-minimal/qwiklabs.yml
@@ -1,7 +1,6 @@
 entity_type: Quiz
 schema_version: 1
 
-id: quiz-minimal
 passing_percentage: 100
 default_locale: en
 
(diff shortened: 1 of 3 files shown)


joncass:
I agree that we should remove top-level `id`, but don't we still need `id`s for the other things for the locale merging?

jgan:
Re: @joncass 's comment about managing locales.

I think that depends on whether the English and Japanese translation of a question the "same" question from the standpoint of reporting? If so, then we need an ID; otherwise we can get away without one.

I'll defer to whoever is making final product decisions about quizzes, but... it seems reasonable to me, though I hope people won't use it to generate silly reports with misleading data.

@rickcnagy - Let's say we do need an ID. We could keep the ID here in the bundle spec, but have it be auto-generated as a content hash in GitWhisperer. That way we don't put the burden on authors to maintain unique IDs.

PS - I'm confident we can manage the l10n workflow without an author-specified ID by using the same hashing mechanism. That way we don't need to require it in the authoring environment just to satisfy l10n operational requirements.

joncass:
@jgan what do you want to do with this PR?

PR #25, opened 2019-03-20
Remove extra 'id' fields from quiz/qwiklabs.yaml

davetchen:
The top-level 'id' (slug) is passed in by Git Whisper. The per-item ones aren't used.


davetchen:
How about if the item `id`s are generated by Git Whisperer (based on the yaml file ordering) if they don't already exist so the field could be "optional"?

joncass:
The ids are there because we specifically do not trust the yaml file ordering. See this comment from @jgan: https://github.com/CloudVLab/git-whisperer/commit/fa4d9a7b167c085b8d8cd2407ea903c97f566ed8#diff-6e21db65baddec4deaed64bf79d12ce4R37

PR #28, opened 2019-04-19
Changes for dm_template[:script]

davetchen:
This field needs to be a directory rather than a single file.

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -162,7 +162,9 @@ fleet                         |          | enum*   | Specify a Qwiklabs fleet to
   id: secondary_project
   fleet: gcpfree
   dm_template:
-    script: deployment_manager
+    script:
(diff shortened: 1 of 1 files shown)


lab-bundle-spec.md
```diff
@@ -162,7 +162,7 @@ fleet                         |          | enum*   | Specify a Qwiklabs fleet to
   id: secondary_project
   fleet: gcpfree
   dm_template:
-    script: deployment_manager/instance_pool.yaml
+    script: deployment_manager
```
akefeli: I find these 2 examples confusing because one has locales under script, one doesn't

akefeli: Also existing labs do not use 'dm_template' at all. They all use 'dm_script' field

davetchen:
Fixed the `dm_template.script` error

PR #29, opened 2019-05-23
Update lab-bundle-spec.md

davetchen:
Add `:display` settings to labs bundle spec

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -26,7 +26,13 @@ description:
 duration: 60
 level: intro
 tags: [sample, life-changing, gcp]
-display: [connection_fleetconsole, connection_vnc, connection_rdp, connection_custom, connection_access_key_id]
+legacy_display_options: [
+  hide_connection_fleetconsole,
+  show_connection_vnc,
+  show_connection_rdp,
+  show_connection_custom,
+  show_connection_access_key_id
+]
 
 ...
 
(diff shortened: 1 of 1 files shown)


lab-bundle-spec.md
```diff
@@ -26,6 +26,7 @@ description:
 duration: 60
 level: intro
 tags: [sample, life-changing, gcp]
+display: [connection_fleetconsole, connection_vnc, connection_rdp, connection_custom, connection_access_key_id]
```
jgan: I like that the array format is flexible, but I have two concerns:

1. These settings have different default values, i.e. `display_connection_fleetconsole` is true by default in the current logic. This is an optional property, so mismatched default behavior is maintenance headache (and I'd argue confusing to authors).

2. I'd like to explicitly call out that this is "legacy" behavior that we intend to move into per-resource configuration... even if the ETA is at least quarters away.

How do you feel about something like...

```
legacy_display_options: [
  hide_connection_fleetconsole,
  show_connection_vnc,
  show_connection_rdp
]
```

Here the default behavior is clear and each option is expressed as an explicit override. This is unfortunately more verbose, but I'd argue the additional clarity will benefit authors and make the un-bundling logic much easier to maintain.

davetchen: Sounds good! Looks like only the first one is default true

davetchen:
@jgan change as requested!

PR #31, opened 2019-07-15
Delete the examples, which are being moved to google3.

dseomn:
These files are being moved to google3 in http://cl/258195375


akefeli:
There could be authors still referencing these. I'd refrain from deleting them until we have the alternative

dseomn:
> There could be authors still referencing these. I'd refrain from deleting them until we have the alternative

https://critique.corp.google.com/#review/258195375 is in your queue with auto submit enabled, so that's up to you now :)

joncass:
@akefeli @dseomn can you please resolve this PR?

akefeli:
let's keep these here for now, since people still update it. but I believe the future for this repo it should be in g3 and have all the spec markdowns auto generated from schemas.

PR #34, opened 2019-11-11
V2 lab bundle spec

joncass:


lab-bundle-spec.md
```diff
@@ -199,10 +207,12 @@ fleet                         |          | enum*   | Specify a Qwiklabs fleet to
 attribute   | required | type       | notes
 ----------- | -------- | ---------- | ----------------------------------------
 permissions |          | array      | Array of project/roles(array) pairs
+ssh_project |          | string     | If an SSH key is desired for thise user, the id of the project that the SSH key should be attached to.
```
joncass: I don't understand the SSH stuff that well, so this may or may not make sense. Looking for feedback on this.

lab-bundle-spec.md
```diff
@@ -158,21 +153,34 @@ environment_resources:
...
-dm_template.custom_properties |          | array   | Array of key/value pairs.
-fleet                         |          | enum*   | Specify a Qwiklabs fleet to pull the project from.
+attribute                           | required | type    | notes
+----------------------------------- | -------- | ------- | --------------------------------------
+dm_template.start.script            |          | path    | Relative path to a Deployment Manager directory tree.
+dm_template.start.custom_properties |          | array   | Array of key/value pairs.
+dm_template.end.script              |          | path    | Relative path to a Deployment Manager directory tree.
+dm_template.end.custom_properties   |          | array   | Array of key/value pairs.
+fleet                               |          | enum*   | Specify a Qwiklabs fleet to pull the project from.
+service_account_permissions         |          | array   | Array of project/roles(array) pairs to be granted to the default compute engine service account in this project.
```
joncass: `service_account_permissions` was the best name I could come up with. Open to other suggestions.

lab-bundle-spec.md
dcbrandao: Navy was recently updated such that an SSH key is now created for each gaia user and the related information is now stored on the website side, so the user flag here is probably no longer necessary.
On the other side, the project needs to specify which user's SSH key should be associated with the project

joncass: Do we actually need to create an ssh key for each user though? Or could we just do it when specified?
As for the project side, I figured if we had the user side indicate which project to attach to, then we could extract that into the project details at unbundling time by finding the project with the id in question.

dcbrandao: Navy currently creates an SSH key for each user, so there's no additional overhead on the website side. this was done in order to cut down on the number of GCP API calls needed during lab launch 
Multiple projects could use (and if they use any, it's likely they will use) the same default SSH key

joncass: Ah gotcha. Okay - updated in https://github.com/CloudVLab/qwiklabs-content-bundle-spec/pull/34/commits/e8a574e90992f14d50cb7fb5cfbda423ddb92e86.

lab-bundle-spec.md
```diff
@@ -158,21 +153,35 @@ environment_resources:
...
   dm_template:
-    script: deployment_manager
-    custom_properties:
-      - key: userNameWindows
-        value: student
+    start:
+      script: dm_startup.zip
+      custom_properties:
+        - key: userNameWindows
+          value: student
```
dcbrandao: we need to figure out how to allow these values to reference properties/outputs from other resources (the authoring side part of #4081)

lab-bundle-spec.md
```diff
@@ -141,15 +141,10 @@ id        |          | string | Identifier that can be used throughout project b
 environment_resources:
   - type: gcp_project
     id: my_primary_project
-    dm_template:
-      script: deployment_manager
+  - type: gcp_project
+    id: secondary_project
   - type: gcp_user
     id: primary_user
-    permissions:
```
dcbrandao: what was the motivation behind removing this section?

lab-bundle-spec.md
```diff
@@ -158,21 +153,35 @@ environment_resources:
...
 ##### GCP Project (gcp_project)
 
-attribute                     | required | type    | notes
------------------------------ | -------- | ------- | --------------------------------------
-dm_template.script            |          | path    | Relative path to a Deployment Manager directory tree.
-dm_template.custom_properties |          | array   | Array of key/value pairs.
-fleet                         |          | enum*   | Specify a Qwiklabs fleet to pull the project from.
+attribute                           | required | type    | notes
+----------------------------------- | -------- | ------- | --------------------------------------
+dm_template.start.script            |          | path    | Relative path to a Deployment Manager directory tree.
```
dcbrandao: this hierarchy (dm_template -> start / end -> script / custom_properties) feels a little awkward to me

lab-bundle-spec.md
```diff
@@ -158,21 +153,35 @@ environment_resources:
...
+dm_template.start.custom_properties |          | array   | Array of key/value pairs.
+dm_template.end.script              |          | path    | Relative path to a Deployment Manager directory tree.
+dm_template.end.custom_properties   |          | array   | Array of key/value pairs.
+fleet                               |          | enum*   | Specify a Qwiklabs fleet to pull the project from.
```
dcbrandao: variant is more or less the replacement for fleet (since resource providers are built on top of navy, not on top of the fleet model), and i believe variant should be a property on all resource types.
this does mean that labs that use weird fleets aren't compatible with the new resource model

lab-bundle-spec.md
joncass: would you prefer something like
```
startup_template:
  script: ...
  custom_properties: ...
cleanup_template:
  script: ...
  custom_properties: ...
```
?

lab-bundle-spec.md
joncass: I felt like it was weird to have this example include permissions but not any of the other resource-specific stuff defined below. Since this is the example right below the type/id definition, I thought it made sense for it to just have types and ids.

lab-bundle-spec.md
joncass: ah right. I think it makes the most sense for custom_properties to have a `key` and EITHER a `value` (string that is passed through) OR `reference` (in the form `[RESOURCE].[PROPERTY]`) rather than try to smuggle both through as `value`. Does that seem reasonable to you? Also, do you have a list of the possible properties per resource?

lab-bundle-spec.md
joncass: we can add `variant` as an optional property for all resources. can you say more about how that makes weird-fleet-labs incompatible? and is that a trade-off you're ok with?

lab-bundle-spec.md
jgan: I do prefer the latter.

`startup_script` and `cleanup_script` as top-level/first-class properties emphasize the role it plays on the lab life-cycle. The fact that it uses deployment manager is an implementation detail that could change.

Note: not wedded to "script" as the noun but I prefer it to "template")

Note: My argument assumes that the clean-up and start-up scripts actually work independently of each other (e.g. don't rely on a shared DM state)... I hope this is the case, but I might be missing a key implementation concern.

lab-bundle-spec.md
dcbrandao: it's actually relying on navy that makes weird-fleet-labs incompatible, because navy doesn't support weird fleets. there's a separate thread of work on the navy side on figuring out what to do with weird fleets

lab-bundle-spec.md
dcbrandao: i think this is more of a question for @jgan. i was original in favor of that, but he had mentioned using some reference scheme like is done in deployment manager.

the reasoning behind having references was to have a replacement for the default list of inputs we pass to deployment manager (lib/ice/gcp/deploymentmanager_v2.rb#161). we may want to expand the list of properties that resources can have, but the ones currently implemented are:
- user
  - password
  - username
- project
  - project_id
  - gcp_zone (maybe should just be 'zone' to match existing dm input name)

currently not accessible through project (but passed in through default dm properties):
  - region
  - key_file_json
  - qwiklabsUuid

jgan: I had assumed we were going to use an interpolation syntax, a la [Terraform](https://www.terraform.io/docs/configuration-0-11/interpolation.html). But, after mulling it for a bit, I'm for the more explicit distinction. It will be simpler to implement and validate.

The primary drawback I see is the lack of ability to do any value manipulation before getting passed to the script.

1) Simple value manipulation (e.g. concat)

```yaml
custom_properties:
  - key: userNameWindows
  - value: "${user.name}-pre-defenestration"
```

I'm not concerned about this, because the target scripting language can handle these value manipulations.

2) Combining an arbitrary number of values, e.g. to construct a list

```yaml
custom_properties:
  - key: listOfUsers
  - value: "[ ${user1.username}, ${user2.username}, ${user3.username} ]"
```

This only effects the scripts re-usability. The script could be re-worked to accept N different custom_properties instead of an array. I suspect this will have a larger impact on clean-up scripts than start-up scripts.

__Net-Net:__ These arguments against the explicit `reference` syntax smell like YAGNI to me. Furthermore, the reference syntax can be expanded to handle more complex usecases (e.g. lists).

lab-bundle-spec.md
jgan: I'm mildly concerned about the ambiguity in the name, since there's a bunch of service accounts involved in the lab runtime. It's more verbose, but how about...

- `gce_service_account_permissions`,
- `default_service_account_permissions`, or
- `default_gce_service_account_permissions`?

lab-bundle-spec.md
jgan: We should indicate the type of script somewhere. Whether that's expressed a `type` field or a nested object, I'm agnostic.

lab-bundle-spec.md
joncass: Updated

lab-bundle-spec-v1.md
```diff
@@ -0,0 +1,248 @@
+# Qwiklabs Lab Bundle Specification
```
supersam654: Is any part of this file up for review in this PR or is this just here for reference/because files moved around?

joncass: The latter

joncass: Well I guess this file's existence and name would be up for discussion here:
- Should we keep the v1 spec in the repo?
- Should it go in a separate folder of "old specs" or stay at the top-level
etc

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
+
+###### Custom properties
+
+attribute | required | type   | notes
+----------| -------- | ------ | --------------------------------------
+key       | ✓        | string | How the property will be referenced within the script.
+value     |          | string | A value to be passed into the script.
+reference |          | string | A reference to a value, in the form `[RESOURCE].[PROPERTY]`, which will be obtained from the running resource and passed into the script.
+
+The valid `reference`s are:
```
jgan: TODO: Let's add a "Referenceable Outputs/Properties(?)" to each resource type description, instead of centralizing them here.

lab-bundle-spec-v1.md
supersam654: I'm fine leaving it around. After v2 is finalized, we should probably add a message to the top of the file asking people to use v2 and also mention that we're deprecating v1 in q7.

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
+- gcpfree
+- gcpasl
+
+
+###### Custom properties
+
+attribute | required | type   | notes
+----------| -------- | ------ | --------------------------------------
+key       | ✓        | string | How the property will be referenced within the script.
+value     |          | string | A value to be passed into the script.
```
supersam654: Please clarify how this field is supposed to be used. Is this a fallback value? A constant to pass along at a later point?  Something else entirely?

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
+reference |          | string | A reference to a value, in the form `[RESOURCE].[PROPERTY]`, which will be obtained from the running resource and passed into the script.
+
+The valid `reference`s are:
+- [USER_ID].username
+- [USER_ID].password
+- [PROJECT_ID].project_id
```
supersam654: Nit: How about `[GCP_PROJECT]` or just `[PROJECT]` instead?

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
+- [PROJECT_ID].zone
```
supersam654: Nit: how about `default_zone` to draw more attention to the fact that this is the default place a student will create resources but is by no means the only place they can do so?

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
+###### Custom properties
+
+attribute | required | type   | notes
+----------| -------- | ------ | --------------------------------------
+key       | ✓        | string | How the property will be referenced within the script.
+value     |          | string | A value to be passed into the script.
+reference |          | string | A reference to a value, in the form `[RESOURCE].[PROPERTY]`, which will be obtained from the running resource and passed into the script.
+
+The valid `reference`s are:
+- [USER_ID].username
```
supersam654: Nit: How about `[GAIA_USER]` or just `[USER]`?

lab-bundle-spec.md
```diff
@@ -158,41 +163,63 @@ environment_resources:
...
-> - gcpfree
-> - gcpasl
->
-> Future versions of the Content Bundle spec may use different terminology for
-> resource_type variations to avoid conflation.
+> Some fleet types map to resource types (e.g. `gsuite_multi_tenant` fleet is
+> now the `gsuite-domain` resource type), while other fleets are allowed as
+> "variants" of `gcp_project` (see below).
+
+###### Variants
```
supersam654: I think variant is a forward-compatible abstraction for fleets but I no longer think it's the complete picture.

I think a project is defined as:
* billing account
* quotas
* APIs

I think variant properly encapsulates all valid (billing_account, APIs) pairs. I'm on the fence on splitting those right now. It would open the door to help reduce our API enablement volume on certain APIs (like ML) but Navy is nowhere near practically supporting arbitrary lists of APIs.

I foresee "quotas" being defined as `cpu_count`, `gpu_count`, `tpu_count`, and `egress_volume`. Given that we don't support customizing any of those right now and adding those fields are all compatible with the current v2 spec, I'd recommend leaving them out entirely until we're ready to seriously support them.

With all of that said, what do you think about replacing variant with:

| variant | billing_account | cpu_count |
| - | - | - |
| gcpd | _default_ | _default_ |
| gcpfree | none | _default |
| gcpondemand | _default_ | 9 |
| gcpedu | edu | _default_ |

joncass: So to clarify, you're saying there would be no `fleet` or `variant` property. Instead, we would add two properties to the `gcp_project` resource: `billing_account` and `cpu_count`. Is that correct?

If so, I think that has some merit. My thoughts are:
- Decoupling the `billing_account` and `cpu_count` could reduce combinatorial explosion as we unlock more values for these. On the flip side, it prevents us from silently changing these params (e.g. uniting or splitting billing accounts, changing gcpondemand to use 12 cpus, etc). I don't know how likely either of these scenarios (adding values or silently changing them) is though.
- I liked the idea that `variant` would be an attribute that all resource types could have. That goes away in the new proposal.
- I don't like adding more fields that have a seemingly-random set of allowed values.
- I do like that this has obvious extensions when we do allow customizing gpus, tpus, egress volume.

So, tl;dr: I see pros and cons and don't have a strong opinion either way. Would be very curious to hear from others (in particular @jgan and @dcbrandao)

lab-bundle-spec.md
joncass: Good question. Now that I'm looking at it again, I'm not sure why this value couldn't just be hardcoded in the script. @dcbrandao ?

lab-bundle-spec.md
joncass: Done

lab-bundle-spec.md
joncass: Done

lab-bundle-spec.md
joncass: Done

lab-bundle-spec.md
joncass: Done

lab-bundle-spec-v1.md
joncass: Agreed

lab-bundle-spec.md
dcbrandao: i believe we (josh, greg, and i) had a conversation about this sort of thing when we were working on the structure of navy, and we decided that we'd be okay with having variants(/subtypes) that corresponded to each of the combinations of those things that we wished to support

supersam654: I think modern Navy subtypes (which are the same thing as variants) encapsulate supported (billing_account, apis, cpu_count). Going forward, I'm concerned we're going to need to separate out `cpu_count` to help with abuse checking and separate out `apis` to appease certain, problematic GCP services.

Sooo...I think we should split `variant` into a few different fields to support other changes we'd like to eventually make in runtime.

dcbrandao: because (type, subtype) is so fundamental to navy, what you're suggesting sounds like a major overhaul of the navy backend, unless what you're really intending is we retain the current type identifier but have some mapping from (billing_account, apis, cpu_count) into more subtypes.

that being said, with respect to the content bundle spec, i think listing out the used apis is fine. this is something the author has complete control over, and we can use it immediately for other purposes.
cpu_count seems problematic. we don't currently give authors enough tools to determine what the cpu count should be, i haven't been convinced that cpu count is the proper metric we should be using for abuse checking, and i'm not sure that authors should necessarily need to care about abuse checking
are you thinking of billing_account as some enum where we interpret what the value is supposed to mean, or as some gcp identifier?

lab-bundle-spec.md
jgan: I'm _very slightly_ concerned about the meaning of "default" here.

There are two "defaults" in operation here:

1. The default zone of the `Lab` as specified by the lab author.
2. The default zone of the `LabInstance` as determined by the context. For instance, a classroom in Japan can override the default zone of the lab to use a closer region/zone.

.... but, unlike the Lab model in the website code base, ContentBundle does not support lab level `default_zone`. So either the point is moot or we need to re-add it.

lab-bundle-spec.md
supersam654: Derek and I chatted and I think we agreed on:

* `billing_account` is an enum of (none, edu, asl, or default)
* `cpu_count` is an optional integer that defaults to the billing account default value. It's generally 6. If a creator wants `gcpondemand`, set this field to 9.
* `apis` is an optional list of APIs that must be enabled in the project. It defaults to the billing account default value which is ~85 APIs for gcpd. For now, we don't have a way to technically do anything with this list of APIs but we'd like to in the future.

@dcbrandao please double check this.

dcbrandao: i'm not sure we agreed on cpu_count, i'm still not sold on this manner of specifying abuse checking

lab-bundle-spec.md
supersam654: Relatedly, I'm concerned that people will try taking a lab in a zone that does not support all of the features they need. Practically, 90% of labs happen in us-central1 which is one of the first places to get new features so this hasn't come up much in the past.

See https://cloud.google.com/about/locations/?tab=americas for a breakdown of features by region. But, long story short, it is not always safe to attempt a lab in a region that is geographically close to the lab taker.

Instead, what if labs specified a list of `valid_regions` and Core Platform handled picking a region that was close to the lab taker? Not specifying a list would imply the lab is available in all regions.

lab-bundle-spec.md
jgan: Currently, authors use custom properties primarily for script re-usability. The same DM script can be used in multiple labs with variable tweaks.

You can check out the `custom_dm_properties` table in Gravitron to get a taste for how their used.

lab-bundle-spec.md
supersam654: `billing_account` without `cpu_count` can't distinguish `gcpd` from `gcpondemand`. Also, here's the design doc I was talking about that isn't doing too well in review :) https://docs.google.com/document/d/1HVYxroXhWkaCaOGpg7vSS_Kg7g2LkoRWdcyy54gdFbQ/edit#

lab-bundle-spec.md
```diff
@@ -210,6 +239,12 @@ permissions |          | array      | Array of project/roles(array) pairs
           - roles/bigquery.admin
 ```
 
+###### Valid custom property references
+
+The valid `reference`s for the `gcp_project` resource are:
```
supersam654: Nit: `gcp_project` -> `gcp_user`

lab-bundle-spec.md
```diff
@@ -135,21 +143,18 @@ The properties of each environment resource will depend on their type, i.e. AWS
...
+    id: secondary_project
+    variant: gcpfree
   - type: gcp_user
     id: primary_user
-    permissions:
-      - project: my_primary_project
-        roles:
-          - roles/editor
-          - roles/bigquery.admin
   - type: gcp_user
```
supersam654: Nit: I know we haven't really talked about it yet, but "GCP user" is a misnomer and this resource is really a Gaia User. Given that v2 includes breaking changes from V1, I'd like to rename this to `gaia_user`.

To make upgrading easier, we can also start supporting `gaia_user` in the V1 spec and then doa  large find+replace on all V1 content so creators don't have to think about this when upgrading to V2.

lab-bundle-spec-v1.md
```diff
@@ -0,0 +1,248 @@
...
+
+##### Future Resource Types
+
+- AWS Account (aws-account)
+- iPython Notebook (ipython-notebook)
+
+> **NOTE:** A draft of the `aws-account` resource type was previously specified
+> in this document. See [previous version](https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/93896ced4ae5b543132d7a10d838ac17bd5ae3e1/lab-bundle-spec.md) for details.
+
+### Activity Tracking (Alpha)
```
supersam654: Do we want to better codify this in V2 or is that a huge task that is beyond the scope of this PR?

lab-bundle-spec.md
joncass: But they still need to copy the file into their lab folder, right? Why wouldn't they just make the property replacement at that time?

lab-bundle-spec-v1.md
joncass: Unless someone is well on their way to figuring out what this would look like for V2, I think we need to punt it to the next version.

lab-bundle-spec.md
joncass: Done

lab-bundle-spec.md
joncass: I'm fine with changing this to `gaia_user` - @jgan and @dcbrandao sound ok to you?

dcbrandao: technically sam is correct, but 'gaia user' is a super confusing term. if i google for that, i get directed to a registration page for an online game

davetchen: agree with @dcbrandao

lab-bundle-spec.md
davetchen: @dcbrandao not sure how you would specify abuse checking then? Seems like the 2 choices are by amount of resources used (which correlates to project charge) or by checking the students behavior with those resources. Even if the student was performing the lab steps as intended we wouldn't want them to over use resources

dcbrandao: well, i have some questions about this in general:
 - do we want authors involved in the abuse enforcement (i.e., by specifying what should be considered abuse on the lab)?
 - if so, is cpu count the right metric to give to the lab author wrt abuse enforcement, or is there some other metric or combination of metrics? 
 - should abuse enforcement happen per-resource (which is what cpu count on a given gcp project would be) or across the entire lab?

supersam654:
My 2 cents on those questions are yes, yes, and per-resource. Happy to chat
more when I'm back in CAM though.

lab-bundle-spec-v1.md
jgan: Unfortunately, the latter.

lab-bundle-spec.md
joncass: @supersam654 are you ok with leaving it as `gcp_user` given Derek and Dave's comments?

lab-bundle-spec.md
joncass: ping @jgan @dcbrandao

lab-bundle-spec.md
joncass: I think there are three options for "should authors be involved in abuse enforcement":
1. Yes, they should have complete control and specify numerical thresholds for all abuse metrics
2. Yes, we should give them a handful of variants that they can pick between that bundle up abuse metrics (status quo)
3. No, runtime should adaptively identify the expected resource usage and tune abuse detection accordingly

I would like to be moving towards (3) in the long-term, but for now I prefer (2). Especially because we can bundle other considerations into the variants to help the authors decide / nudge them in the right direction. Note that if go with (1), the author has no incentive to pick values other than the max for everything.

lab-bundle-spec.md
```diff
@@ -1,18 +1,26 @@
...
+- Adding an `ssh_key_user` attribute on the `gcp_project` resource type.
+
 ## `qwiklabs.yaml` Structure
 
 Here's a sample `qwiklabs.yaml` file, with all nested details removed to make it easier to see the general file structure.
 
 ```yml
 entity_type: Lab
-schema_version: 1
+schema_version: 2
```
joncass: Note that @jgan is working on creating a PR on top of this PR to:
- Include AWS considerations
- Rework `legacy_display_options`

lab-bundle-spec.md
supersam654: I don't feel super strongly either way so `gcp_user` is fine. With that said, I'd at least like a comment explaining why we're sticking with the misnomer.

lab-bundle-spec.md
supersam654: Fascinating! I like (2) for now because it strikes a good balance between effort and reward but I'd like to move towards (1) over time for greater flexibility and letting people lock down labs that would otherwise be too dangerous to use.

As for your point on authors not having the right incentives, I agree as of today but I'm hoping that is not the case long term (cue my rickety design doc on lab locking).

lab-bundle-spec.md
joncass: Added "The `gcp_user` type could more properly be called `gaia_user`, since that's what it provisions. However, the term `gaia` is less well-known, so we stick with `gcp`." below

lab-bundle-spec.md
joncass: We can definitely discuss which direction we want to move, but it sounds like we're agreed that for now we should stick with (2), so I think this issue is resolved.

lab-bundle-spec.md
joncass: It's a constant to be passed in, which allows a template to be reused across labs without having to laboriously replace all instances of one word with another when copying it around. It will also make it easier to enable true template reuse in the future.

lab-bundle-spec.md
supersam654: In practice, does the v2 bundle spec need to support `legacy_display_options`?

joncass: No it does not. Josh's PR will be removing it.

PR #35, opened 2019-11-22
V2 display options

jgan:
Add `display_options` to the lab environment.


lab-bundle-spec.md
```diff
@@ -264,6 +277,34 @@ No additional attributes
 > **NOTE:** A draft of the `aws-account` resource type was previously specified
 > in this document. See [previous version](https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/93896ced4ae5b543132d7a10d838ac17bd5ae3e1/lab-bundle-spec.md) for details.
 
+#### Display Outputs
```
jgan: @reviewers  - I'm not a huge fan of the name "Display Outputs". I'm trying to evoke something closer to the idea of "Givens" in a problem set.

Any suggestions would be appreciated.

joncass: Some ideas:
- student_facing
- provided
- surfaced
- public

lab-bundle-spec.md
```diff
@@ -264,6 +277,34 @@ No additional attributes
...
+
+Specify which resources properties are given to the lab taker.
```
joncass:
```suggestion
Specify which resource properties are given to the lab taker.
```

lab-bundle-spec.md
```diff
@@ -264,6 +277,34 @@ No additional attributes
...
+
+Not all details of the lab environment should be exposed to the lab taken. For example, a lab may involve a GCP project and two GCP users. The lab taker is expected to log into GCP as one user, and manipulate the IAM privileges of the other. Since the lab taker is not expected to log in to the second user, there is no reason to display the GCP user's password and doing so may be distract.
```
joncass:
```suggestion
Not all details of the lab environment should be exposed to the lab taker. For example, a lab may involve a GCP project and two GCP users. The lab taker is expected to log into GCP as one user, and manipulate the IAM privileges of the other. Since the lab taker is not expected to log in as the second user, there is no reason to display the second user's password and doing so may be distracting.
```

lab-bundle-spec.md
```diff
@@ -264,6 +277,34 @@ No additional attributes
...
+    - label:
+        locales:
+          en: "Username"
+      reference: primary_user.username
+    - label:
+        locales:
+          en: "Password"
+      reference: primary_user.password
+```
+
```
joncass: add something like "See each resource type's "Valid custom property references" section for the valid references."?

jgan: I like it. :+1: 

```suggestion
See each resource type's "Valid custom property references" section for the valid references.
```

lab-bundle-spec.md
jgan: I really like 'student_facing', because it's clear about who's perspective it's taking.

But I worried about 'facing' being to idiomatic, so I went with 'student_visible'. If anyone thinks I'm being too sensitive, let me know. At a gut level I definitely prefer 'student_facing'.

PR #38, opened 2019-12-18
Add AWS account resource type

jgan:


lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+default_policy.allow_aws_subnet_deletion     |          | boolean | Default false.
+default_policy.allow_aws_vpc_deletion        |          | boolean | Default false.
+startup_script.type              |          | string  | The type of startup script. Only `cloud_formation` is supported.
+startup_script.path              |          | path    | Relative path to a directory tree with the script contents.
+startup_script.custom_properties |          | array   | Array of pairs. See [Custom Script Properties](#custom-script-properties) for details.
+cleanup_script.type              |          | string  | The type of cleanup script. Only `cloud_formation` is supported.
+cleanup_script.path              |          | path    | Relative path to a directory tree with the script contents.
+cleanup_script.custom_properties |          | array   | Array of pairs. See [Custom Script Properties](#custom-script-properties) for details.
+user_policy                      |          | path    | Relative path to a [JSON policy](https://awspolicygen.s3.amazonaws.com/policygen.html) document
+variant                          |          | enum*   | Specify a Qwiklabs fleet name.
```
joncass: `variant` is already specified as a property of all resources, so I think this line can be removed

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+  default_policy:
+    allow_aws_dedicated_instances: false
+    allow_aws_ondemand_instances: false
+    allow_aws_rds_instances: ['db.t2.micro']
+    allow_aws_spot_instances: false
+    allow_aws_subnet_deletion: false
+    allow_aws_vpc_deletion: false
+```
+
+###### Valid Variants for AWS Account
```
joncass: please make this more consistent with the `Variants` section in GCP Project

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
   id: primary_domain
 ```
 
-##### Future Resource Types
+##### AWS Account (aws_account)
+
+attribute   | required | type       | notes
+----------- | -------- | ---------- | ----------------------------------------
+default_policy.allow_aws_dedicated_instances |          | boolean | Default false.
+default_policy.allow_aws_ondemand_instances  |          | array   | Array of [EC2 instance types](#valid-eC2-instance-types) that are valid Ondemand usage. Default none.
```
joncass:
```suggestion
default_policy.allow_aws_ondemand_instances  |          | array   | Array of [EC2 instance types](#valid-eC2-instance-types) that are valid for Ondemand usage. Default none.
```
what does "Ondemand" mean in this context?

joncass: also, I find it confusing that most of the `allow_` things are booleans, but this and the next one are arrays. should they instead be `allowed_`?

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+  id: the_account
+  variant: sts
+  startup_script:
+    type: cloud_formation
+    path: ./lab.template
+    custom_properties:
+      - key: userNameWindows
+        value: student
+  user_policy: ./student.policy
+  default_policy:
```
joncass: why are these under `default_policy`? i know that's a term from AWS infra, but not sure how useful it is here. also, should/could we have a single `policies` header that accounts for the below and the `user_policy`?

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+- aws_ec2
+- aws_rt53labs_ilt
+- aws_sts
+- aws_vpc
+- awsiam
+
+... or ....
+
+List of all fleets as currently used:
+https://console.cloud.google.com/bigquery?sq=403355294977:d397a1416639404e9b320ec78f476209
```
joncass: i'm assuming this is a just a note for us to decide what variants to specify, but just to clarify: most people will not be able to run this query

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+
+
+###### Valid EC2 Instance Types
+
+The list of valid EC2 instance types is not fixed.
+
+TODO(joshgan): Find link for evergreen reference to valid content types.
+
+TODO(joshgan): Should Qwiklabs/Alexandria validate this at time of ingestion/publishing?
```
joncass: my instinct is no

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+
+###### AWS Account Default Policy
```
joncass: i would remove this section if we are documenting above all the default_policy options that we support. of we are allowing a full policy document to be specified, then it would make sense to delete the above fields and have more info here. i think.

lab-bundle-spec.md
```diff
@@ -271,13 +269,92 @@ No additional attributes
...
+default_policy.allow_aws_spot_instances      |          | boolean | Default false.
+default_policy.allow_aws_subnet_deletion     |          | boolean | Default false.
+default_policy.allow_aws_vpc_deletion        |          | boolean | Default false.
+startup_script.type              |          | string  | The type of startup script. Only `cloud_formation` is supported.
+startup_script.path              |          | path    | Relative path to a directory tree with the script contents.
+startup_script.custom_properties |          | array   | Array of pairs. See [Custom Script Properties](#custom-script-properties) for details.
+cleanup_script.type              |          | string  | The type of cleanup script. Only `cloud_formation` is supported.
+cleanup_script.path              |          | path    | Relative path to a directory tree with the script contents.
+cleanup_script.custom_properties |          | array   | Array of pairs. See [Custom Script Properties](#custom-script-properties) for details.
+user_policy                      |          | path    | Relative path to a [JSON policy](https://awspolicygen.s3.amazonaws.com/policygen.html) document
```
joncass:
```suggestion
user_policy                      |          | path    | Relative path to a [JSON policy](https://awspolicygen.s3.amazonaws.com/policygen.html) document.
```

lab-bundle-spec.md
jgan:
```suggestion
```

lab-bundle-spec.md
jgan: Good question. I'll add some more documentation about what this section actually does. But to answer you directly, there are few important clarifying points:

`default_policy` can be thought of as the account level policies. 
The `default_policy` input is not structured as a normal policy file. These values are flags that Qwiklabs uses to construct policy.
Lab authors are not allowed to directly specify this policy, because there are a number of properties that Qwiklabs needs to ensure are always set (e.g. settings that prevent abusers from taking full control of the account and breaking the cleaning cycle).

lab-bundle-spec.md
jgan: Yup, this is just for review purposes. I don't expect this to be committed in the final version.

lab-bundle-spec.md
joncass: Maybe rename it to `account_restrictions` or something, so people don't get confused with the full `default_policy` concept? Or `default_policy_flags` or `default_policy_args`?

lab-bundle-spec.md
jgan: Do you mean just make the title "Variants"?

lab-bundle-spec.md
jgan: Oooh... I like both `account_restrictions` and `default_policy_flags`. I'll go with `account_restrictions` because it is a bit more flexible in usage; but I'll happily exchange it for the alternative.

lab-bundle-spec.md
joncass: yes

lab-bundle-spec.md
```diff
@@ -271,13 +269,90 @@ No additional attributes
...
+```
+
+###### Variants for AWS Account
+
+- aws_beta
+- aws_ec2
+- aws_rt53labs_ilt
+- aws_sts
+- aws_vpc
+- awsiam
```
dcbrandao: in the current method of launching aws accounts, 'awsiam' and 'awsfed' are separate fleet types. which of those should be aws_account, and are they similar enough that they should be considered variants or should they be different resource types? (cc @eniskonuk)

lab-bundle-spec.md
```diff
@@ -271,13 +269,90 @@ No additional attributes
...
+    custom_properties:
+      - key: userNameWindows
+        value: student
+  user_policy: ./student.policy
+  account_restrictions:
+    allow_dedicated_instances: false
+    allow_spot_instances: false
+    allow_subnet_deletion: false
+    allow_vpc_deletion: false
+    allowed_ondemand_instances: []
```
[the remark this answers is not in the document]
jgan: In code: yes.
In practice: 🤷‍♂

- [ ] TODO(joshgan): Answer this in Gravitron

lab-bundle-spec.md
```diff
@@ -271,13 +269,90 @@ No additional attributes
...
-##### Future Resource Types
+##### AWS Account (aws_account)
+
+attribute   | required | type       | notes
+----------- | -------- | ---------- | ----------------------------------------
+account_restrictions.allow_dedicated_instances  |          | boolean | Default false.
+account_restrictions.allow_spot_instances       |          | boolean | Default false.
+account_restrictions.allow_subnet_deletion      |          | boolean | Default false.
+account_restrictions.allow_vpc_deletion         |          | boolean | Default false.
+account_restrictions.allowed_ondemand_instances |          | array   | Array of [EC2 instance types](#valid-eC2-instance-types) that are valid for Ondemand usage. Default none.
```
jgan: - [ ] TODO(joshgan): renamed to `allowed_ec2_instances`

lab-bundle-spec.md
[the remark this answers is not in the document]
jgan: That is equivalent. We have changed the name accordingly.

lab-bundle-spec.md
```diff
@@ -271,13 +269,90 @@ No additional attributes
...
+  id: the_account
+  variant: sts
+  startup_script:
+    type: cloud_formation
+    path: ./lab.template
+    custom_properties:
+      - key: userNameWindows
+        value: student
+  user_policy: ./student.policy
+  account_restrictions:
```
dgorelik: Sorry for the drive-by comment, but I was talking to the AWS CE folks and they were hoping to be able to specify a maximum number of managed compute (Lightsail, SageMaker, EBS) instances. Should that be added here?

jgan: When/if we add the functionality, that's something we can add here.

dgorelik: So, broadly speaking, we are free to add fields to this spec when we want?

lab-bundle-spec.md
jgan: :+1:

lab-bundle-spec.md
jgan: Broadly speaking, yes.

jgan: As long as it's backwards compatible (which additive changes usually are)

jgan:
> What about some more uncommon fields that are used in CloudFormation:
> 
> - [x] `lab.image_id_param`

There are 5 uses of this on AWS staging, and they all have titles that are some variation of "Test Lab".

> - [x] `lab.ignore_stack_errors`

This is actually a property of lab_instance, and it's passed in as a parameter to `lab.run`. Quick link: https://github.com/CloudVLab/qwiklab-website/blob/add-index-on-catalog-subdomain-filter/app/models/lab.rb#L1107

> - [x] `lab.cfn_create_stack_tags`

This is a derived property. Quick link: https://github.com/CloudVLab/qwiklab-website/blob/add-index-on-catalog-subdomain-filter/app/models/lab.rb#L975:L975

> - [ ] `lab.qwiklab_id`

I thought this was only used by the Nvidia use case... but I'm seeing some AWS and GCP uses. I need to take a deeper look.

> - [x] `lab.global_set_vnc_password`

This isn't set directly by authors. It's value depends on `SysSetting.cur.lab_options_force_set_vnc_password`

jgan:
@supersam654 

> What about some more uncommon fields that are used in CloudFormation:
> 
> - [x] `lab.image_id_param`

There are 5 uses of this on AWS staging, and they all have titles that are some variation of "Test Lab".

> - [x] `lab.ignore_stack_errors`

This is actually a property of lab_instance, and it's passed in as a parameter to `lab.run`. Quick link: https://github.com/CloudVLab/qwiklab-website/blob/add-index-on-catalog-subdomain-filter/app/models/lab.rb#L1107

> - [x] `lab.cfn_create_stack_tags`

This is a derived property. Quick link: https://github.com/CloudVLab/qwiklab-website/blob/add-index-on-catalog-subdomain-filter/app/models/lab.rb#L975:L975

> - [ ] `lab.qwiklab_id`

I thought this was only used by the Nvidia use case... but I'm seeing some AWS and GCP uses. I need to take a deeper look.

> - [x] `lab.global_set_vnc_password`

This isn't set directly by authors. It's value depends on `SysSetting.cur.lab_options_force_set_vnc_password`

jgan:
Follow-up re: `lab.qwiklab_id`

The data leads me to believe it the lab edit UI would set a default value equal to the current deployment. I believe we can safely ignore it in this spec. I'm merging this PR, but will happily accept a change if we discover we need it.

PR #39, opened 2019-12-19
v2 activity tracking spec

joncass:
Add a spec for multi-resource activity tracking.

Of note:
- `manual`:
  - Omits `notify` (used 100 times out of 200,000) and declared legacy by [QL AT deck](https://docs.google.com/presentation/d/1H-di-0GzzSZNWWXyIg1oq29_3SX-fW6pR75L4Bqg7oU)
  -  Omits `score_completion_only` (used 160 times out of 200,000)
  - Omits `manual`, `manual_remote_url`, which are deprecated
  - Omits `name`, `description` (not sure what these would be for at the Manual level)
- `step`:
  - Allows title to be localized
  - Allows student-facing messages to be localized
  - Allows services from multiple resources
  - Omits `api` (since there might be multiple environment types)
  - Omits `max`, `interval` (QL code should be deciding these things [or, preferably IMO, removing the background checks altogether])
  - Omits `method_name`, and assumes the method to be called will be named `check` (other methods can still be created, and will not be called by the QL code)
  - Omits `multi_region` (this has caused lots of headaches and is only used 700 times out of 875,000)
- `step.code`:
  - Does not take `points` as an input (not sure why this would be useful)
  - Does not allow omitting a non-student message
  - Does not output a `done` key (will instead be inferred from `score == maximum_score`
  - Is still just a chunk of ruby code, rather than allowing for other languages or references to common modules

TODO: 
- add specification of allowed services in the various resource type sections
- update the `examples` directory

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -333,7 +333,7 @@ attribute        | required | type                       | notes
 -----------------| -------- | ---------------------------| --------------------------------------
 title            | ✓        | locale dictionary          |
 maximum_score    | ✓        | integer                    | The maximum number of points this step can award.
-student_messages | ✓        | array of objects           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
+student_messages | ✓        | dictionary of locale dictionaries           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
 services         | ✓        | array of resource services | An array of services that will be used in the code block. Each resource type specifies a set of allowed services.
 code             | ✓        | string                     | Code to be executed. See [below](#code) for more information.
 
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -368,7 +368,7 @@ assessment:
       services:
         - target_project.StorageV1
       code: |-
-        def check(handles, max_score)
+        def check(handles: handles, resources: resources, maximum_score: maximum_score)
           storage_handle = handles['target_project.StorageV1']
 
           # Check for bucket
@@ -383,7 +383,7 @@ assessment:
             return { score: 2, student_message: 'bucket_misconfigured' }
           end
 
-          { score: max_score, student_message: 'success' }
+          { score: maximum_score, student_message: 'success' }
         end
     - title:
         locales:
@@ -413,7 +413,7 @@ assessment:
         - source_project.StorageV1
         - target_project.StorageV1
       code: |-
-        def check(handles, max_score)
+        def check(handles: handles, resources: resources, maximum_score: maximum_score)
           target_storage_handle = handles['target_project.StorageV1']
           source_storage_handle = handles['source_project.StorageV1']
 
@@ -430,7 +430,7 @@ assessment:
             return { score: 2, student_message: 'file_mismatch' }
           end
 
-          { score: max_score, student_message: 'success' }
+          { score: maximum_score, student_message: 'success' }
         end
 ```
 
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -374,16 +374,16 @@ assessment:
           # Check for bucket
           found_bucket = ...
           unless found_bucket
-            return { score: 0, message: 'bucket_missing' }
+            return { score: 0, student_message: 'bucket_missing' }
           end
 
           # Check bucket configuration
           bucket_configured_correctly = ...
           unless bucket_configured_correctly
-            return { score: 2, message: 'bucket_misconfigured' }
+            return { score: 2, student_message: 'bucket_misconfigured' }
           end
 
-          { score: max_score, message: 'success' }
+          { score: max_score, student_message: 'success' }
         end
     - title:
         locales:
@@ -420,17 +420,17 @@ assessment:
           # Check for file
           found_file = ...
           unless found_file
-            return { score: 0, message: 'file_missing' }
+            return { score: 0, student_message: 'file_missing' }
           end
 
           # Check file contents
           source_contents = ...
           target_contents = ...
           unless source_contents == target_contents
-            return { score: 2, message: 'file_mismatch' }
+            return { score: 2, student_message: 'file_mismatch' }
           end
 
-          { score: max_score, message: 'success' }
+          { score: max_score, student_message: 'success' }
         end
 ```
 
@@ -444,4 +444,4 @@ The method `check` will be called with two arguments:
 
 The method `check` should return a single hash with:
 - `:score`: the number of points the student earned.
-- `:message`: a key from the step's `student_messages` array, which will be presented to the student in the appropriate locale.
+- `:student_message`: a key from the step's `student_messages` array, which will be presented to the student in the appropriate locale.
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -368,7 +368,7 @@ assessment:
       services:
         - target_project.StorageV1
       code: |-
-        def check(handles)
+        def check(handles, max_score)
           storage_handle = handles['target_project.StorageV1']
 
           # Check for bucket
@@ -383,7 +383,7 @@ assessment:
             return { score: 2, message: 'bucket_misconfigured' }
           end
 
-          { score: 5, message: 'success' }
+          { score: max_score, message: 'success' }
         end
     - title:
         locales:
@@ -413,7 +413,7 @@ assessment:
         - source_project.StorageV1
         - target_project.StorageV1
       code: |-
-        def check(handles)
+        def check(handles, max_score)
           target_storage_handle = handles['target_project.StorageV1']
           source_storage_handle = handles['source_project.StorageV1']
 
@@ -430,7 +430,7 @@ assessment:
             return { score: 2, message: 'file_mismatch' }
           end
 
-          { score: 5, message: 'success' }
+          { score: max_score, message: 'success' }
         end
 ```
 
@@ -438,8 +438,9 @@ assessment:
 
 The code block must be valid Ruby code. It must have a method called `check`, and may optionally contain helper methods as well.
 
-The method `check` will be called with a single argument:
+The method `check` will be called with two arguments:
 - A map from `[RESOURCE].[SERVICE]` to a handle to that resource's service. It will only contain handles to services specified in the step's `services` array.
+- The maximum score for the step.
 
 The method `check` should return a single hash with:
 - `:score`: the number of points the student earned.
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -308,20 +308,88 @@ environment:
       reference: primary_user.password
 ```
 
-### Activity Tracking (Alpha)
+### Activity Tracking
 
 Activity tracking is a feature for evaluating a student's performance in a lab by running a script at "checkpoints". These scripts can call APIs relevant to any environment resource to query their current state. For example, the script may inspect and validate the configuration of GCE instances running in `my-project`, to ensure the user is following the instructions properly.
 
-Lab bundles will provisionally support the JSON representation of Activity Tracking currently used in the Qwiklabs web interface. The JSON definition should be stored in file separately from (and referenced directly in) `qwiklabs.yaml`.
+A lab has a Manual, which in turn contains the Steps (checkpoints).
 
-```yaml
-entity_type: Lab
+#### Manual
 
-...
+attribute          | required | type    | notes
+-------------------| -------- | --------| --------------------------------------
+passing_percentage | ✓        | integer | The percentage of total points the student must achieve to "pass" the lab.
+steps              | ✓        | array   | An array of [Steps](#steps)
 
-activity_tracking: ./assessment/activity_tracking.json
+```yml
+manual:
+  passing_percentage: 75
+  steps: ...
 ```
 
-> **Note:** Support for this format should be considered deprecated.
->
-> Further work will be done to define a new DSL for expressing Activity Tracking logic. Exploratory sketches of this DSL can be found in the `./examples` directory of this repository.
+#### Steps
+
+attribute        | required | type                       | notes
+-----------------| -------- | ---------------------------| --------------------------------------
+title            | ✓        | locale dictionary          |
+maximum_score    | ✓        | integer                    | The maximum number of points this step can award.
+student_messages | ✓        | array of objects           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
+services         | ✓        | array of resource services | An array of services that will be used in the code block. Each resource type specifies a set of allowed services.
+code             | ✓        | string                     | Code to be executed. See [below](#code) for more information.
+
+```yml
+manual:
+  passing_percentage: 75
+  steps:
+    - title:
+      en: Create a Cloud Storage bucket
+      es: Crear un depósito de almacenamiento en la nube
+    maximum_score: 5
+    student_messages:
+      - success:
+          en:
+            Great job! You created the bucket!
+          es:
+            ¡Gran trabajo! ¡Creaste el cubo!
+      - bucket_missing:
+          en:
+            Oops! No bucket found.
+          es:
+            ¡Uy! No se ha encontrado el cubo.
+      - bucket_misconfigured:
+          en:
+            Hmm. The bucket is there, but it is misconfigured.
+          es:
+            Hmm. El cubo está allí, pero está mal configurado.
+    services:
+      - my_gcp_project.StorageV1
+    code: |-
+      def check(handles)
+        storage_handle = handles['my_gcp_project.StorageV1']
+        
+        # Check for bucket
+        found_bucket = ...
+        unless found_bucket
+          return { score: 0, message: 'bucket_missing' }
+        end
+        
+        # Check bucket configuration
+        bucket_configured_correctly = ...
+        unless bucket_configured_correctly
+          return { score: 2, message: 'bucket_misconfigured' }
+        end
+        
+        { score: 5, message: 'success' }
+      end
+```
+
+##### Code
+
+The code block must be valid Ruby code. It must have a method called `check`, and may optionally contain helper methods as well.
+
+The method `check` will be called with a single argument:
+- A map from `[RESOURCE].[SERVICE]` to a handle to that resource's service. It will only contain handles to services specified in the step's `services` array.
+
+The method `check` should return a single hash with:
+- `:score`: the number of points the student earned.
+- `:message`: a key from the step's `student_messages` array, which will be presented to the student in the appropriate locale.
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -312,9 +312,9 @@ environment:
 
 Activity tracking is a feature for evaluating a student's performance in a lab by running a script at "checkpoints". These scripts can call APIs relevant to any environment resource to query their current state. For example, the script may inspect and validate the configuration of GCE instances running in `my-project`, to ensure the user is following the instructions properly.
 
-A lab has a Manual, which in turn contains the Steps (checkpoints).
+A lab has an Assessment, which in turn contains the Steps (checkpoints).
 
-#### Manual
+#### Assessment
 
 attribute          | required | type    | notes
 -------------------| -------- | --------| --------------------------------------
@@ -322,7 +322,7 @@ passing_percentage | ✓        | integer | The percentage of total points the s
 steps              | ✓        | array   | An array of [Steps](#steps)
 
 ```yml
-manual:
+assessment:
   passing_percentage: 75
   steps: ...
 ```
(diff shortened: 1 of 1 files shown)


jgan:
> * `manual`:

I know this corresponds directly with our ActiveRecord model name, but I've never understood the noun "manual" in this context.

Suggested alternatives: `assessment`, `evaluation`

> * `step`:
>   
>   * Allows title to be localized
>   * Allows student-facing messages to be localized

I'm 10,000% in favor of this, but it does imply database model and lab UI changes before we could have authors use the format.

>   * Omits `max`, `interval` (QL code should be deciding these things [or, preferably IMO, removing the background checks altogether])

(Agreed [and agreed])

>   * Omits `method_name`, and assumes the method to be called will be named `check` (other methods can still be created, and will not be called by the QL code)

I have mixed feelings about this. I'd like to give advanced authors the ability to write activity tracking in separate files for shared functionality, easier linting, and possibly unit testing prior to Test Harness. In this case, they would likely need to identify which function in the file(s) applied to which step.... but (1) YANGI, and (2) we could give them an alternate specification format for references to other files.

> * `step.code`:
>   
>   * Does not take `points` as an input (not sure why this would be useful)

I just poked around and it seems to be used to decouple the code from the max score. So if the max score for a step changes (excitement factor!), author doesn't need to remember to change it in both the metadata and the code.

I don't feel strongly either way; but if we keep it, then I'd suggest we at least start calling the parameter `max_points`/`max_score`.

>   * Does not allow omitting a non-student message

:clap: :clap: :clap: 

>   * Does not output a `done` key (will instead be inferred from `score == maximum_score`

I'm cool with this because in practice that's how it's always used. I searched the `gcp-spl-content` repo and found that every usage of `done => true` corresponded with returning the maximum score.

However, as an _entirely new feature_ in the future, I wonder if the concept of "I'm done, do not allow re-evaluation" is interesting to authors.

>   * Is still just a chunk of ruby code, rather than allowing for other languages or references to common modules

:+1:

joncass:
@jgan thanks for the comments! Responses below.

> > * `manual`:
> 
> I know this corresponds directly with our ActiveRecord model name, but I've never understood the noun "manual" in this context.
> 
> Suggested alternatives: `assessment`, `evaluation`
> 
Updated `manual` -> `assessment`

> > * `step`:
> >   
> >   * Allows title to be localized
> >   * Allows student-facing messages to be localized
> 
> I'm 10,000% in favor of this, but it does imply database model and lab UI changes before we could have authors use the format.
> 
Yes, there will need to be changes in `website`, but that was going to be the case anyway.

> > * Omits `method_name`, and assumes the method to be called will be named `check` (other methods can still be created, and will not be called by the QL code)
> 
> I have mixed feelings about this. I'd like to give advanced authors the ability to write activity tracking in separate files for shared functionality, easier linting, and possibly unit testing prior to Test Harness. In this case, they would likely need to identify which function in the file(s) applied to which step.... but (1) YANGI [sic], and (2) we could give them an alternate specification format for references to other files.
> 
I agree that in theory an author may want to specify a method name, but in practice I don't see that happening and we can definitely add an optional param, or other syntax entirely if/when that comes up.

> > * `step.code`:
> >   
> >   * Does not take `points` as an input (not sure why this would be useful)
> 
> I just poked around and it seems to be used to decouple the code from the max score. So if the max score for a step changes (excitement factor!), author doesn't need to remember to change it in both the metadata and the code.
> 
> I don't feel strongly either way; but if we keep it, then I'd suggest we at least start calling the parameter `max_points`/`max_score`.
> 
The decoupling makes sense, but I doubt that max scores change very often, and would prefer to keep it simpler. I could definitely be swayed on this one though.

> > * Does not output a `done` key (will instead be inferred from `score == maximum_score`
> 
> I'm cool with this because in practice that's how it's always used. I searched the `gcp-spl-content` repo and found that every usage of `done => true` corresponded with returning the maximum score.
> 
> However, as an _entirely new feature_ in the future, I wonder if the concept of "I'm done, do not allow re-evaluation" is interesting to authors.
> 
Something to consider for the future.

joncass:
After conversation with Enis, to make things more backwards-compatible, I added the `max_score` param back in, and renamed the output message to `student_message`

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+              en:
+                Hmm. There's a file here, but it doesn't match the source contents.
+              es:
+                Hmm. Hay un archivo aquí, pero no coincide con el contenido de origen.
+      services:
+        - source_project.StorageV1
+        - target_project.StorageV1
+      code: |-
+        def check(handles, max_score)
+          target_storage_handle = handles['target_project.StorageV1']
```
dcbrandao: do we want to pass handles in this format, or more like ```handles[:target_project][:StorageV1]```?
also we will probably need to list the possible services for each resource type

davetchen:
Hi Jon, thanks for taking this on! Seems like a couple old "features" have been removed -- just want to make sure that's intentional:

- notify (browser/OS notifications)
- score_completion_only (pass/fail)

And in "steps"
- interval (step must be completed within a certain time)
- multi_region (explicitly support multi-region check for the step)
- api (e.g. GCP -- perhaps can always be inferred through runtime context??)

lab-bundle-spec.md
joncass: i'd be fine with passing handles in either format - what do you think?
and yeah I have it as a "TODO" in the PR description to enumerate the services for each resource type.

joncass:
@davetchen yeah, please see the description of the PR - I think I addressed all of those.

lab-bundle-spec.md
supersam654: Is changing the contract of the arguments to this method in scope for a v2 bundle change? Alternatively, could we make a metadata change for all of the other fields and then talk about how we want the code snippet to function?

joncass: yes, changing the contract of the arguments is in scope - that's actually the main reason for taking this on now: that multi-resource labs necessarily need a different signature

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+                Oops! No bucket found.
+              es:
+                ¡Uy! No se ha encontrado el cubo.
+        - bucket_misconfigured:
+            locales:
+              en:
+                Hmm. The bucket is there, but it is misconfigured.
+              es:
+                Hmm. El cubo está allí, pero está mal configurado.
+      services:
```
supersam654: I'd like creators to specify a per-step field and I don't know what to call it but basically I'd like creators to specify what granularity their step applies to:

* `zone`: Run this in every GCP zone (e.g. checking if a GCE instance was created in that zone)
* `region`: Run this in every GCP region (e.g. checking if a regional GKE cluster was made in that region)
* `multiregion`: Run this once for US, EU, and Asia (e.g. checking for multi-regional GCS buckets or some newer BigQuery data locality stuff)
* `global`: Run this function once (e.g. checking IAM configuration)

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+
+```yml
+assessment:
+  passing_percentage: 75
+  steps:
+    - title:
+        locales:
+          en: Create a Cloud Storage bucket
+          es: Crear un depósito de almacenamiento en la nube
+      maximum_score: 5
```
supersam654: I know you removed the pass/fail checkbox based on its lack of usage but I'm curious how many steps offer partial points.

Instead of only allowing integer scores and passing percentages, I'd like to propose only allowing pass/fail steps and a passing step percentage (which defaults to all of them in Git Whisperer).

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+                ¡Uy! No se ha encontrado el cubo.
+        - bucket_misconfigured:
+            locales:
+              en:
+                Hmm. The bucket is there, but it is misconfigured.
+              es:
+                Hmm. El cubo está allí, pero está mal configurado.
+      services:
+        - target_project.StorageV1
+      code: |-
```
supersam654: I'd like to also include a `type` or `code_type` field that indicates this step specifies its code as a Ruby method with particular arguments. GitWhisperer could infer it during bundling until we support more types but I think that would make it easier to extend this particular part without messing with the rest of the metadata (and revving the spec version) in the future.

lab-bundle-spec.md
joncass: interesting. i don't think we want to run any of this stuff in every zone, and even running it in every region has caused us headaches. i would much rather keep everything "global" until we have a really compelling use case and/or rearchitect things to make "run n times" much easier.

lab-bundle-spec.md
joncass: I'm open to that if others are strongly in favor of it, but my preference would be to leave it out for now. If/when it is useful, we can add a field for that in a backwards-compatible way (i.e. without changing the spec version) by just assuming `ruby` if not specified.

lab-bundle-spec.md
joncass: it's not just a question of how many steps offer partial points - this would also change things if a given assessment has steps with different `max`s. it would also change how game points are calculated. i think i'd rather leave as-is for now, unless others feel strongly.

eniskonuk:
> Add a spec for multi-resource activity tracking.
> 
> Of note:
> 
> * `manual`:
>   
>   * Omits `notify` (used 100 times out of 200,000) and declared legacy by [QL AT deck](https://docs.google.com/presentation/d/1H-di-0GzzSZNWWXyIg1oq29_3SX-fW6pR75L4Bqg7oU)

This is fine. An experiment that should have been cleaned up a while ago.

>   * Omits `score_completion_only` (used 160 times out of 200,000)

It is OK, I believe, to remove this from the assessment spec but we do need something like this as an attribute of the lab. It is more of a UX thing. There are some labs where the content creator doesn't want to display scores but a checkbox for completing/passing a step. The assessment code should always work with scores.

>   * Omits `manual`, `manual_remote_url`, which are deprecated
>   * Omits `name`, `description` (not sure what these would be for at the Manual level)

This is also fine. It was unclear when we originally designed this whether these would be useful fields or not. It is time to remove.

> * `step`:
>   
>   * Allows title to be localized
>   * Allows student-facing messages to be localized
>   * Allows services from multiple resources

This is awesome and very much needed.

>   * Omits `api` (since there might be multiple environment types)

Not sure I understand what this is.

>   * Omits `max`, `interval` (QL code should be deciding these things [or, preferably IMO, removing the background checks altogether])

This was also an experiment that didn't work but I'm not 100% whether something like this is needed in the future or not. The thinking here was that in a lab with multiple steps, students would have a portion of the lab time allocated to completing a particular step and if they didn't, they would fail that step even if they complete it later in the lab. The implementation of this ended up being too confusing but I can see a requirement like this for actual challenge labs/tests/games.

>   * Omits `method_name`, and assumes the method to be called will be named `check` (other methods can still be created, and will not be called by the QL code)

This is OK as long as we are always running the method in complete isolation but wouldn't this also make it harder for us to implement global variables in the future like what Harish as requested. To be sure, I am against that for the moment but I can see us doing something like that in the more distant future.

>   * Omits `multi_region` (this has caused lots of headaches and is only used 700 times out of 875,000)

I hate this as well and I would like to also remove it but we still don't have a good solution for students that don't do the lab in the region/zone the lab is supposed to run in. I'm not sure it's OK for us to remove this without having an answer to that question.

> * `step.code`:
>   
>   * Does not take `points` as an input (not sure why this would be useful)

This changed based on backward compatibility right?

>   * Does not allow omitting a non-student message

Not sure I understand what this is.

>   * Does not output a `done` key (will instead be inferred from `score == maximum_score`

I am not 100% sure of this one. Partial credit is not used at the moment but not because it's not useful but because it is beyond the capabilities of the creators. So, if we are saying, the only way to do partial credit is by creating a lot of simple steps and not a single complex (multiple evaluation) step I am OK with that but would like to at least paper test the premise of it. If 'we end up putting this back in again in the near future I would then just keep it.

Also, this creates a backwards compatibility issue potentially that we need to solve for.

>   * Is still just a chunk of ruby code, rather than allowing for other languages or references to common modules

Yup. That is fine.
> 
> TODO:
> 
> * add specification of allowed services in the various resource type sections
> * update the `examples` directory

joncass:
@eniskonuk thanks for the comments. My responses below.

> > * Omits `score_completion_only` (used 160 times out of 200,000)
> 
> It is OK, I believe, to remove this from the assessment spec but we do need something like this as an attribute of the lab. It is more of a UX thing. There are some labs where the content creator doesn't want to display scores but a checkbox for completing/passing a step. The assessment code should always work with scores.
> 
I don't even know if it should be a lab attribute, given how rarely it is used. For now we can at least remove it from the AT spec, and we can do some research on the use cases before we decide how/if to add it back to the platform.

> > * Omits `api` (since there might be multiple environment types)
> 
> Not sure I understand what this is.
> 
The current `api` field is one of `AWS`, `GCP`, `GSuite`, which determines what handles are available in the step. That does not make sense in a multi-environment lab.

> > * Omits `max`, `interval` (QL code should be deciding these things [or, preferably IMO, removing the background checks altogether])
> 
> This was also an experiment that didn't work but I'm not 100% whether something like this is needed in the future or not. The thinking here was that in a lab with multiple steps, students would have a portion of the lab time allocated to completing a particular step and if they didn't, they would fail that step even if they complete it later in the lab. The implementation of this ended up being too confusing but I can see a requirement like this for actual challenge labs/tests/games.
> 
Ah interesting. I think we should remove it here, and if we have similar use cases in the future we can design something that meets the requirements at that time.

> > * Omits `method_name`, and assumes the method to be called will be named `check` (other methods can still be created, and will not be called by the QL code)
> 
> This is OK as long as we are always running the method in complete isolation but wouldn't this also make it harder for us to implement global variables in the future like what Harish as requested. To be sure, I am against that for the moment but I can see us doing something like that in the more distant future.
> 
I think it would actually make it easier for us to add globals in the future, since we wouldn't have to worry about conflicting with arbitrary method names that authors come up with.

> > * Omits `multi_region` (this has caused lots of headaches and is only used 700 times out of 875,000)
> 
> I hate this as well and I would like to also remove it but we still don't have a good solution for students that don't do the lab in the region/zone the lab is supposed to run in. I'm not sure it's OK for us to remove this without having an answer to that question.
> 
Given the exceptionally low usage of `multi_region`, it seems like the vast majority of labs are getting by just fine without accounting for students working in unexpected region/zone. I don't know if there's something special about the labs that are using this, or if it's random, but I think until we hear more about what makes this a key feature, we should remove it. Especially in a multi-environment lab, `multi_region` as currently implemented is untenable.

> > * `step.code`:
> >   
> >   * Does not take `points` as an input (not sure why this would be useful)
> 
> This changed based on backward compatibility right?
> 

Correct

> > * Does not allow omitting a non-student message
> 
> Not sure I understand what this is.
> 

Activity tracking currently allows the return hash to have two "message" entries: `message` and `student_message`. I think the theory was that `student_message` would be conveyed back to the student, while `message` is logged, but I don't think it's really used.

> > * Does not output a `done` key (will instead be inferred from `score == maximum_score`
> 
> I am not 100% sure of this one. Partial credit is not used at the moment but not because it's not useful but because it is beyond the capabilities of the creators. So, if we are saying, the only way to do partial credit is by creating a lot of simple steps and not a single complex (multiple evaluation) step I am OK with that but would like to at least paper test the premise of it. If 'we end up putting this back in again in the near future I would then just keep it.
> 
> Also, this creates a backwards compatibility issue potentially that we need to solve for.
> 

You can still do partial credit without the `done` key. The current setup would have something like:
- `max_score` is 5, you return `{ points: 2, done: false }`, which means that the student got 2/5 and has not "completed" the step
- `max_score` is 5, you return `{ points: 5, done: true }`, which means that the student got 5/5 and has  "completed" the step
- `max_score` is 5, you return `{ points: 2, done: true }`, which means that the student got 2/5 and *has* "completed" the step, despite not getting all the points

I am proposing that you just return `{ points: 2 }` or `{ points: 5 }` and we would know that in the first case you are not done, while in the second you are. Does that make sense?

I also don't think it creates a backwards compatibility issue, since we can just ignore the `done` key in the return.

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
-> **Note:** Support for this format should be considered deprecated.
->
-> Further work will be done to define a new DSL for expressing Activity Tracking logic. Exploratory sketches of this DSL can be found in the `./examples` directory of this repository.
+#### Steps
+
+attribute        | required | type                       | notes
+-----------------| -------- | ---------------------------| --------------------------------------
+title            | ✓        | locale dictionary          |
+maximum_score    | ✓        | integer                    | The maximum number of points this step can award.
+student_messages | ✓        | array of objects           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
```
dcbrandao: i think this should be a dictionary of objects instead of an array of objects

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+            locales:
+              en:
+                Hmm. There's a file here, but it doesn't match the source contents.
+              es:
+                Hmm. Hay un archivo aquí, pero no coincide con el contenido de origen.
+      services:
+        - source_project.StorageV1
+        - target_project.StorageV1
+      code: |-
+        def check(handles, max_score)
```
dcbrandao: something that the current activity tracking code runner does is augment these handles with :projects and :users for GCP api and :domain and :users for GSuite api. we need to figure out how we want to provide these for v2 lab activity tracking, as we are no longer providing an API field

lab-bundle-spec.md
jgan: Agreed

lab-bundle-spec.md
jgan: TODO: Agreed. Why not both?

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+
+The code block must be valid Ruby code. It must have a method called `check`, and may optionally contain helper methods as well.
+
+The method `check` will be called with two arguments:
+- A map from `[RESOURCE].[SERVICE]` to a handle to that resource's service. It will only contain handles to services specified in the step's `services` array.
+- The maximum score for the step.
+
+The method `check` should return a single hash with:
+- `:score`: the number of points the student earned.
+- `:student_message`: a key from the step's `student_messages` array, which will be presented to the student in the appropriate locale.
```
jgan: - [ ] TODO(joshgan): Student message should be passed through if it's not found in the locale hash.
- [ ] TODO(joshgan): Do we need to both `message` and `student_message`. It's unclear how `message` is currently used.

lab-bundle-spec.md
steve-rivera: Is there a reason the key is `maximum_score`, but in the code example below we pass in `max_score`?

lab-bundle-spec.md
jgan, line 336:
```suggestion
student_messages | ✓        | dictionary of locale dictionaries           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
```

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+            locales:
+              en:
+                Hmm. There's a file here, but it doesn't match the source contents.
+              es:
+                Hmm. Hay un archivo aquí, pero no coincide con el contenido de origen.
+      services:
+        - source_project.StorageV1
+        - target_project.StorageV1
+      code: |-
+        def check(handles, max_score)
```
jgan, line 416: @joncass, @supersam654 

```suggestion
        def check(handles:handles, resources: resources, maximum_score: maximum_score)
```

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+        - bucket_misconfigured:
+            locales:
+              en:
+                Hmm. The bucket is there, but it is misconfigured.
+              es:
+                Hmm. El cubo está allí, pero está mal configurado.
+      services:
+        - target_project.StorageV1
+      code: |-
+        def check(handles, max_score)
```
jgan, line 371: @joncass @supersam654 

```suggestion
        def check(handles:handles, resources: resources, maximum_score: maximum_score)
```

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+            return { score: 0, student_message: 'bucket_missing' }
+          end
+
+          # Check bucket configuration
+          bucket_configured_correctly = ...
+          unless bucket_configured_correctly
+            return { score: 2, student_message: 'bucket_misconfigured' }
+          end
+
+          { score: max_score, student_message: 'success' }
```
joncass, line 386:
```suggestion
          { score: maximum_score, student_message: 'success' }
```

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+          end
+
+          # Check file contents
+          source_contents = ...
+          target_contents = ...
+          unless source_contents == target_contents
+            return { score: 2, student_message: 'file_mismatch' }
+          end
+
+          { score: max_score, student_message: 'success' }
```
joncass, line 433:
```suggestion
          { score: maximum_score, student_message: 'success' }
```

lab-bundle-spec.md
joncass: in the code example it used to be a positional param so it didn't matter, but consistency would probably have been nice. it's now a keyword param and therefore does matter, and has been updated accordingly.

lab-bundle-spec.md
```diff
@@ -308,20 +308,140 @@ environment:
...
+        end
+```
+
+##### Code
+
+The code block must be valid Ruby code. It must have a method called `check`, and may optionally contain helper methods as well.
+
+The method `check` will be called with two arguments:
+- A map from `[RESOURCE].[SERVICE]` to a handle to that resource's service. It will only contain handles to services specified in the step's `services` array.
+- The maximum score for the step.
```
jgan, line 443:
```suggestion
The method `check` will be called with three keyword arguments:

- `handles`: A map from `[RESOURCE].[SERVICE]` to a handle to that resource's service. It will only contain handles to services specified in the step's `services` array.
- `resources`: A map from `[RESOURCE]` (ids) to maps of each ....
- `maximum_score`: The maximum score for the step.
```

joncass:
At long last, this PR is ready to merge. Thanks all for the comments!

PR #40, opened 2020-01-31
Content Bundle Spec for Exams

carinetorres:


carinetorres:
TBD...
1) randomize_prompts on Exam?
2) retakeable on Exam?

carinetorres:
> Looks good! Will need to add the "retake" fields as that requirement takes shape

Added retake_cooldown

PR #42, opened 2020-02-06
Add a classroom-template bundle spec

davetchen:
First pass at the long awaited bundle spec for ClassroomTemplate

### DIFF
--- a/classroom-template-bundle-spec.md
+++ b/classroom-template-bundle-spec.md
@@ -2,7 +2,10 @@
 
 **Version 1**
 
-> This is a DRAFT document. We welcome feedback as this format evolves.
+> **Note: For Interim Use Only**
+>
+> This specification is for existing Qwiklabs platform users that are currently using the `ClassroomTemplate` model. It should be used to help transition from `ClassroomTemplates` to [`CourseTemplates`](./course-template-bundle-spec.md). If you are not currently using `ClassroomTemplates`, please use `CourseTemplates` instead.
+
 
 ## `qwiklabs.yaml` Structure
 
--- a/classroom-template-bundle-spec.md
+++ b/classroom-template-bundle-spec.md
@@ -84,9 +84,6 @@ max_hot_labs: 30
 # Enable DRM on instructions if supported
 enable_drm: false
 
-# Billing code used for reporting ["gcp", "platform", "aws"]
-billing_code: null
-
 # Check classroom for fraud
 resource_limit_check: false
 
@@ -131,7 +128,6 @@ estimated_duration      |          | integer     | Estimated time to take the co
 lock_position           |          | boolean     | Allow lab order to be changed by the trainer
 max_hot_labs            |          | integer     | Maximum number of hot labs for this course.
 enable_drm              |          | boolean     | Enable DRM on instructions if supported
-billing_code            |          | string      | Billing code used for reporting usually one of [null, "gcp", "platform", "aws"]
 resource_limit_check    |          | boolean     | Check classroom for fraud
 student_resources       |          | array       | Student-specific instructions. See the [Resource Spec](./resource-spec.md) for full specification.
 instructor_resources    |          | array       | Instructor-specific instructions. See the [Resource Spec](./resource-spec.md) for full specification.
--- a/classroom-template-bundle-spec.md
+++ b/classroom-template-bundle-spec.md
@@ -69,9 +69,6 @@ role_tags: ['cloud architect', 'developers backend']
 domain_tags: ['infrastructure']
 level: 1
 
-# Allow classroom scheduling for this ClassroomTemplate
-published: true
-
 # Estimated time to take the course, in days
 estimated_duration_days: 1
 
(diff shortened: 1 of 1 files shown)


davetchen:
Just spoke with Enis and it sounds like `billing_code` is an odd ball field.  The value is set in a `before_create` hook on the staging deployments and then its "passed along" as it moves to production. I'll remove it from the bundle-spec but we'll need to discuss how to get the proper value in production.

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
+# Qwiklabs ClassroomTemplate Bundle Specification
+
+**Version 1**
+
+> This is a DRAFT document. We welcome feedback as this format evolves.
```
jgan: Since this is a public doc, we should explicitly call out that this is pseudo-deprecated and integrators should use CourseTemplates if possible.

```suggestion
> **Note: For Interim Use Only**
>
> This specification is for existing Qwiklabs platform users that are currently using the `ClassroomTemplate` model. It should be used to help transition from `ClassroomTemplates` to [`CourseTemplates`](./course-template-bundle-spec.md). If you are not currently using `ClassroomTemplates`, please use `CourseTemplates` instead.

```

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
...
+version                 |          | string      | Version string used as content decorator
+classroom_type          |          | enum        | One of ["Self-paced", "Bootcamp/Workshop", "Instructor-led"]
+course_code             |          | string      | Id for Course, e.g. T_AHYXXX-I
+title                   | ✓        | dictionary  | A locale dictionary of titles
+description             | ✓        | dictionary  | A locale dictionary of descriptions
+objectives              |          | dictionary  | A locale dictionary of objectives
+audience                |          | dictionary  | A locale dictionary of audiences
+prerequisites           |          | dictionary  | A locale dictionary of prerequisites
+outline                 |          | dictionary  | A locale dictionary of outline elements usually formatted as a JSON string
+on_demand_url           |          | dictionary  | A locale dictionary of URL's to Coursera (or other) on-demand content
```
jgan: - "on-demand" doesn't mean the same thing to everyone, and we're about to start delivering "on-demand" courses on Qwiklabs proper.
- There's more than one of them.

How about `external_delivery_urls`? (I'm not sold on this either)

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
...
+The full specification is as follows:
+
+attribute               | required | type        | notes
+----------------------- | -------- | ----------- | -----------------------------------------
+entity_type             | ✓        | string      | Must be `ClassroomTemplate`
+schema_version          | ✓        | integer     |
+default_locale          | ✓        | enum        | Must be a valid locale code
+version                 |          | string      | Version string used as content decorator
+classroom_type          |          | enum        | One of ["Self-paced", "Bootcamp/Workshop", "Instructor-led"]
+course_code             |          | string      | Id for Course, e.g. T_AHYXXX-I
```
jgan: We might want to put in a note about how this differs from the Alexandria slug.

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
...
+max_hot_labs            |          | integer     | Maximum number of hot labs for this course.
+enable_drm              |          | boolean     | Enable DRM on instructions if supported
+resource_limit_check    |          | boolean     | Check classroom for fraud
+student_resources       |          | array       | Student-specific instructions. See the [Resource Spec](./resource-spec.md) for full specification.
+instructor_resources    |          | array       | Instructor-specific instructions. See the [Resource Spec](./resource-spec.md) for full specification.
+modules                 |          | array       | See below
+
+### Modules
+
+A `ClassroomTemplate` presents labs to the learner in a specific order. To keep some compatibilty with other "bundle entities" we will use a simple collection of steps, each with one lab activity.
```
jgan: :+1:

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
...
+tags                    |          | array       | Array of strings to be used as hints in searching, etc.
+product_tags            |          | array       | Array of strings from the "Products" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+role_tags               |          | array       | Array of strings from the "Roles" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+domain_tags             |          | array       | Array of strings from the "Domain" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+level                   |          | integer     | Integer between 1 and 4, with 1 being the easiest
+estimated_duration_days |          | integer     | Estimated time to take the course, in days
+estimated_duration      |          | integer     | Estimated time to take the course, in minutes
+lock_position           |          | boolean     | Allow lab order to be changed by the trainer
```
jgan: `lock_activity_positions`?

classroom-template-bundle-spec.md
```diff
@@ -0,0 +1,168 @@
...
+    en: "https://www.coursera.org/specializations/architecting-hybrid-cloud-infrastructure-anthos"
+
+tags: [sample, life-changing, gcp]
+product_tags: ['compute engine', 'cloud storage']
+role_tags: ['cloud architect', 'developers backend']
+domain_tags: ['infrastructure']
+level: 1
+
+# Allow classroom scheduling for this ClassroomTemplate
+published: true
```
jgan: Can we nix `published` from the spec and just set it to true if it comes from Alexandria?

classroom-template-bundle-spec.md
truongvi: Will steps have one lab activity or have multiple lab activities like in the example below?

classroom-template-bundle-spec.md
davetchen: Maybe `external_content_url` ?

classroom-template-bundle-spec.md
davetchen: Yep, multiple steps each with one lab activity ... I think that's what the example tries to represent.

davetchen:
@truongvi Hi Victor, made changes as suggested. We'll need to take care of 2 "outlier" cases directly in the unbundling.

- set `billing_code` to one of ["gcp", "platform", "aws"] based on the `library`
- set `published` to `true` in the unbundler

Hopefully that's it!

PR #48, opened 2020-03-11
Add course surveys to classroom template spec

davetchen:
Will have to independently figure out how to publish course surveys (doing it through old publish/subscribe now).

### DIFF
--- a/classroom-template-bundle-spec.md
+++ b/classroom-template-bundle-spec.md
@@ -69,6 +69,10 @@ role_tags: ['cloud architect', 'developers backend']
 domain_tags: ['infrastructure']
 level: 1
 
+# course_surveys
+course_surveys:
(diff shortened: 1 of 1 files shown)


classroom-template-bundle-spec.md
```diff
@@ -69,6 +69,10 @@ role_tags: ['cloud architect', 'developers backend']
 domain_tags: ['infrastructure']
 level: 1
 
+# course_surveys
+course_surveys:
+- 'cloud-training-l2io16cn5i'
```
truongvi: Could this be represented as `course_surveys: ['cloud-training-l2io16cn5i']` because it is just an array of slugs

davetchen: Yep, yaml supports both for arrays -- I can change it if you like!

PR #51, opened 2020-03-25
Make cleanup scripts by invitation only

supersam654:
Per the latest chat between myself, Josh, and Enis, cleanup scripts will exist but we want teams to talk to us before using them.


jgan:
LGTM

PR #52, opened 2020-04-13
Certification bundle spec draft

shoffing:
Corresponding database schema:
Certification Schema (5)


certification-content-bundle-spec.md
```diff
@@ -0,0 +1,83 @@
...
+description        |          | dictionary | A locale dictionary of the certification description which describes the contents of the certification
+steps              | ✓        | array      | An array of `steps` (see [Steps](#steps) below for details)
+
+### Steps
+
+A certification is composed of multiple `steps`. A step can be either a Course Template or an exam that the user must complete to progress through the certification.
+
+#### Gated Steps
+
+Steps can optionally be marked as "gated". A gated step will be unavailable to the user until all preceding steps are completed, and all succeeding steps will be unavailable to the user until the gated step is complete.
```
steve-rivera, line 44: I think gating should only be before the step. So in a 5 step certification, if step 3 is marked gated, once user completes steps 1 and 2, steps 3, 4, and 5 become available.

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,83 @@
...
+course_template_id: my-cool-course
+```
+
+#### Exam Step Example
+
+```yml
+id: step-1
+gated: true
+step_type: exam
+proctoring: ProctorU
```
steve-rivera, line 72: I think there are more proctoring options here. Live vs Recorded and Proctoring Institution. @davetchen maybe you have ideas about how/where that gets represented?

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,83 @@
...
+default_locale: en
+
+title: 
+  locales:
+    en: Virtual Machines in GCP
+
+description:
+  locales:
+    en: Prove what you know about VMs in GCP by doing these courses and exams!
+
```
davetchen, line 24: As part of the basic attributes we should discuss which fields from the sample we want to carry forwards: https://github.com/CloudVLab/qwiklabs-test-content/blob/master/certifications/bec-certificate/qwiklabs.yaml

- objectives
- audience
- prerequisites
- tags
- product_tags
- role_tags
- domain_tags
- credits
- certificate_award

certification-content-bundle-spec.md
davetchen: I guess the question is for the 'gate' step itself. Does it imply a single gate or separate enter/exit gates?

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,83 @@
...
+- Once `C` is completed, `D` becomes available.
+- Once `D` is completed, `E` becomes available.
+
+#### Course Step Example
+
+```yml
+id: step-0
+gated: false
+step_type: course
+course_template_id: my-cool-course
```
davetchen, line 63: I think `gated` by default should be "false" and in this case `step_type` should be "course_template".
Also each step can use a generic `id`. 
I like to include the "library" explicitly e.g. "qwiklabs-test-content/my-cool-course"

certification-content-bundle-spec.md
davetchen: In the sample certification I used "proctor" instead of "proctoring" - not sure which is better?
For now we can stub in two choices
- qwiklabs-live-plus
- qwiklabs-record-plus

certification-content-bundle-spec.md
shoffing: We can't have two `id` fields, keys must be unique in YAML.

How about `content_id` or `ref_id` or something?

And sure, I can default gated to false.

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,118 @@
...
+description        |          | dictionary | A locale dictionary of the certification description which describes the contents of the certification
+objectives         |          | dictionary | A locale dictionary of objective arrays
+audience           |          | dictionary | A locale dictionary of audiences
+prerequisites      |          | dictionary | A locale dictionary of prerequisites
+tags               |          | array      | Array of strings to be used as hints in searching, etc.
+product_tags       |          | array      | Array of strings from the "Products" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+role_tags          |          | array      | Array of strings from the "Roles" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+domain_tags        |          | array      | Array of strings from the "Domain" column in [this sheet](https://docs.google.com/spreadsheets/d/1hUUch85HBRsRJsgRo9VCg0Pn7ZXi21sl6JU7VOr9LP8)
+credits            |          | integer    | Price of the certification
```
joncass, line 72: I'm not sure this belongs here at all: is price an authoring concern? But if it does it should probably be renamed, since `credits` is not a well-known unit.

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,118 @@
...
+certificate_award  | ✓        | string     | A unique identifier for the corresponding award that this certification grants
```
joncass, line 73: Where are awards defined?

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,118 @@
...
+steps              | ✓        | array      | An array of `steps` (see [Steps](#steps) below for details)
+
+### Steps
+
+A certification is composed of multiple `steps`. A step can be either a Course Template or an exam that the user must complete to progress through the certification.
```
joncass, line 78:
```suggestion
A certification is composed of multiple `steps`. A step can be either a [CourseTemplate](./course-template-bundle-spec.md) or an [Exam](./exam-bundle-spec.md) that the user must complete to progress through the certification.
```

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,118 @@
...
+type: course_template
+content_id: qwiklabs-test-content/my-cool-course
+```
+
+#### Exam Step Example
+
+```yml
+id: step-1
+gated: true
+type: course_template
```
joncass, line 107:
```suggestion
type: exam
```

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,118 @@
...
+proctor: qwiklabs-live-plus
+```
+
+attribute          | required | type       | notes
+-------------------| -------- | ---------- | -----------------------------------------
+id                 | ✓        | string     | A unique identifier for this step, to be consistent across revisions
+gated              |          | boolean    | Whether this step should be "gated" (see [Gated Steps](#gated-steps) above for details); `false` by default
+type               | ✓        | enum       | One of ["course_template", "exam"]
+content_id         | ✓        | string     | Unique identifier of this step's corresponding CourseTemplate or Exam
+proctor            |          | enum       | (exam steps only) Which service to use for proctoring this step's corresponding Exam. One of ["qwiklabs-live-plus", "qwiklabs-record-plus"]
```
joncass, line 118: Is this an authoring concern?

certification-content-bundle-spec.md
shoffing: The way I have things now, they would need to be an entirely new and different spec.

certification-content-bundle-spec.md
davetchen: `credits` is used for the Lab price

joncass: which lab price?

shoffing: We've decided that certifications are offerings, and price is to be defined by the author in the bundle spec.

certification-content-bundle-spec.md
shoffing: Yes, the level of proctoring is determined by authors depending on the needs of their certification.

certification-content-bundle-spec.md
shoffing: We've decided to put this off until later. Removing certificate_award from this PR

shoffing: @carinetorres is working on the specifics of this

certification-content-bundle-spec.md
joncass: Yeah, I guess these aren't "authors" in the usual sense - they're really just compiling existing content pieces.

certification-content-bundle-spec.md
joncass: The "level" of proctoring makes sense to me as an authoring concern, but the specific "proctoring service" doesn't

jgan:
@shoffing @davetchen @joncass - Let's hold off on merging this until we've resolved a few outstanding questions.

Merging a PR in this public repo is the equivalent of publishing an API spec, and we're not ready for that.

The fact that this PR is pending should not impede our progress on the implementation in ql website and git whisperer.

shoffing:
Agreed, specifically based on your comment in the models PR it sounds like a new `version` field will be required for the step schema in addition to the `id` (library/slug)

certification-content-bundle-spec.md
jgan: @shoffing and @davetchen - I'm sorry you need to re-litigate this, but I need to hear the rationale for this. Is there a design doc I can read that summarizes the decision?

Just to be clear about why this is a big deal, this is the first bundle type (other than labs) to include price information. Labs included `price` in credits as an expedient for migration. If we add this to Certifications, then we are officially codifying the pattern that _offering_ concerns are valid for _content_ bundles. That may be the right things to do, but there are a number of consequences.

If we're going to include price (in credits) in content bundle spec, then we need to have plans for 

- Who sets the price? Do they have access to the authoring and publishing workflows?
- What is the value of a credit? In USD? What happens if that exchange rate changes in our system? How do we communicate this relationship to authors?
- Can this price be overridden in certain contexts? (e.g. Catalogs)

To be clear, I firmly believe that information in the content bundle spec can _inform_ pricing and offering configurations. 

The simplest implementation could be an MSRP in USD that is displayed when configuring catalog prices (in credits). This MSRP would have no programmatic effect on content purchasing.

A more complicated system for Labs would be us analyze a number of (related) factors, such as average resource usage and abuse risk. The result would inform (at offer time) the minimum cost in credits for lab in any public catalog. We would be free to play with the value of credits and experiment with what those pricing thresholds are to maximize legitimate usage without increased risk.

But given:
(1) we're pretty far away from implementing that system, and
(2) we have high-touch relationships with everyone who creates self-paced lab content.

Therefore, pushing the decision of cost-in-credits as a heuristic onto authors is currently worth the drawbacks to our system overall.

I don't want to take that hit with a new content type that does not significantly benefit from tying itself to credits at authoring/publishing time.

eniskonuk: I am going to stick my nose into this. I don't fully understand the issue here so I am going to ask a bunch of questions:
1- If price is not an authoring concern, then where is it supposed to be set in the Qwiklabs eco-system?
2- If price is a publishing concern, would that change where it should be set?
3- Should one be allowed to change price in a production deployment? Is that what's happening with the Catalog override?
4- How do we set the price for exams if not in the content-bundle? Especially in the short-term?
5- Who is supposed to set the price if it is not the creators?
6- I don't understand why the QL Credit to USD exchange rate has anything to do with this. Can I ask for more detail on why that is a concern?

joncass: Might be worth getting the group (Steve, Dave, Enis, Josh, me, ...?) together for a 30-minute chat.

certification-content-bundle-spec.md
davetchen: We don't expect authors to be defining this, really to define the offering.

joncass: Yeah, I guess I didn't expect us to be using content bundles for offerings. Probably something that warrants further discussion.

certification-content-bundle-spec.md
```diff
@@ -0,0 +1,112 @@
...
+
+- At first, only `A` and `B` are available to the user.
+- Once `A` and `B` are completed (in any order), `C` and `D` become available.
+- Once `C` and `D` are completed (in any order), `E` becomes available.
+
+#### Course Step Example
+
+```yml
+type: course_template
+id: qwiklabs-test-content/my-cool-course
```
davetchen, line 95: @jgan can you please indicate here how the "version" would look?

shoffing: Would it maybe look this?

```
id: qwiklabs-test-content/my-cool-course/1.2.3
```

davetchen: Here's what it looks like coming back from Alexandria

"libraries/gcp-it-cert-content/content/course_1/versions/1.0"

https://www.qwiklabs.com/admin/alexandria_subscriptions/libraries%2Fgcp-it-cert-content/content/libraries%2Fgcp-it-cert-content%2Fcontent%2Fcourse_1%2Fversions%2F1.0

So in general it has major.minor, but we would need to enforce that up and down the chain.

shoffing: Sure, but that rule should be defined and enforced with a different specification / PR. Application-level semantic version format enforcement is beyond the scope of this spec, which is just for defining Certifications. I'll add this expected format here though.

`libraries/gcp-it-cert-content/content/course_1/versions/1.0` is a little verbose. Is there any reason that can't be `gcp-it-cert-content/course_1/1.0`?

davetchen: Just FYI. Confirmed with @akefeli that in Alexandria the "semantic_version" has 0 or 1 periods and that the elements on either side of the period should be a non-negative integer

davetchen: After discussion with @shoffing and @akefeli to make parsing and validation much simpler we decided to separate the version with a different delimiter `@`. So the first `/` denotes the library. Subsequent `/` are part of the slug and the sematic_version is after `@`
For example
```
type: exam
id: qwiklabs-test-content/my-excellent-exam@1.0
```

joncass: I don't think `/` is a valid character for slugs, so we can probably just use slashes for both delimiters. I'd rather not introduce a new convention here that isn't used anywhere else.

shoffing: `/` is a valid character in slugs, which makes it hard to use slashes as both delimiters. Slashes are URL-escaped (CGI) coming from Alexandria, then unescaped in website: https://github.com/CloudVLab/qwiklab-website/blob/master/app/services/alexandria_service.rb#L141-L151

joncass: Huh - that really surprises me. I thought that the whole reason we were using slashes as delimiters in content IDs is because we don't allow slashes in libraries, slugs, etc. Did that change at some point? FWIW it doesn't look like any existing labs have a slash in the slug.

shoffing: This code has been in website since about a year ago, and I don't know about the state of things before it. Here's the PR: https://github.com/CloudVLab/qwiklab-website/pull/3272

joncass: @akefeli @davetchen @jgan do you know why we try to allow slashes in slugs? I was under the impression that we did not allow that. I think we should disallow that and use slashes as delimiters.

shoffing: I just got done chatting with Dave at our standup. So the slashes in slugs are currently being used by git whisperer to categorize the content in directories in git. Authors had a hard time when there were hundreds of labs under a single flat directory. Some authors would use namespacing with dashes or underscores or something (like `steves-cool-labs_goto-is-awesome`) but it wasn't standardized. Allowing them to use slashes (and parsing the slashes into directories) gives them a standardized way to do organization of content which fits well with git.

joncass: Ah gotcha. It makes sense to want to allow labs to be divided into folders, but that just means we need to change our "slug = filesystem location" mapping, not that we should change how our identifiers are formed or delimited.

shoffing: Including the semantic version with the slug and library in identifier strings is entirely new, and does require a change in how the identifier is formed.

Using a new delimiter for the version is the only way to maintain pure backwards compatibility with any existing content. While unlikely to actually exist, a slug of `12345` is currently perfectly valid. If our identifier changes to `[library]/[slug]/[version]` where library and version are optional, the string `thing/12345` becomes impossible to parse. It could mean (library: thing, slug: 12345) or (slug: thing, version: 12345) depending on the unbundler implementation, and neither is inherently more correct.

Using a new delimiter like `@` is explicit, easy to read, easy to understand, and easy to implement. Here's a regex that reliably parses the entire identifier: https://regexr.com/53eem

joncass: I thought there already were examples of `[library]/[slug]/[version]/[revision]` somewhere but I could be making that up. It's definitely something we've talked about before, but maybe we only ended up using `libraries/[library]/content/[slug]/versions/[version]/[revisions]/revision` or whatever. I didn't think we would ever have a case where library and version were both optional, but I suppose that's possible. The other way of doing it then would be to leave `id` as just `[library]/[slug]` and add an extra `version` attribute so that we don't have to keep going with string parsing.

You're right that `@` is explicit, easy to read, easy to understand, and easy to implement - I'm not arguing against any of that. The thing to keep in mind with things like this is whether all of our systems and APIs make the *same* explicit, easy to read, easy to understand, and easy to implement decision. Since the same argument could be made about using `|` or `>` or something else we just need to be careful to make decisions like this publicly and consistently.

If this is truly the first time that we are including the semantic version with slug and library, then I guess any decision is fine, as long as we make sure that the same decision is made when this comes up in a different context.

PR #56, opened 2020-06-02
Add trainer_resource_uri to course_template

joncass:
Add trainer_resource_uri to course_template

I'll update the examples and the related entities (game_template, etc) if people like this change

### DIFF
--- a/course-template-bundle-spec.md
+++ b/course-template-bundle-spec.md
@@ -48,6 +48,10 @@ prerequisites:
   locales:
     en: <p>A basic understanding of serverless architecture is recommended but not required.</p>
 
+trainer_resource_uri:
+  locales:
+    en: http://www.example.com/resource.pdf
+
 tags: [sample, life-changing, gcp]
 product_tags: ['compute engine', 'cloud storage']
 role_tags: ['cloud architect', 'developers backend']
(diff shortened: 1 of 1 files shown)


jgan:
How is this different from `instructor_resources`?

joncass:
Hmm good question. I didn't realize we had `instructor_resources` on course templates. I don't think they're unbundled anywhere. Should we remove `instructor_resources`?

jgan:
Actually, I'd like to do the opposite. Let's start using `instructor_resources` as defined (i.e. an array of resource objects conforming to the other uses resources), instead of having `trainer_resource_uri` as a one-off locale dictionary.

joncass:
Yeah I thought you might say that. I'm not sure what an array of resources means in this case though.

jgan:
Interesting! I'm not sure what an single URL means in this case. :)

What is the use case that caused you to want to add this single link?

Historically, ClassroomTemplates allow for an unbounded number of student and instructor resources. These resources could be videos, pdfs, external links, etc. The only difference between the two is whether students are allowed to see them.

The instructor-only resources are used to hand-off a number of supplementary resources, such as additional notes on topics, marked-up versions of the slides, and guidance about proper timing and pacing of the course.

We certainly don't need to slavishly copy the behavior of ClassroomTemplates for Courses, but having a list of resources for instructors-only is used broadly.

joncass:
Yeah, the design in the instructor dashboard for course templates is to have a master "Train the Trainer" doc with all the notes/info the trainer would want in a single pdf. See the "Before you begin ..." part of https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZcJe6UKbin6f/files/MCEJu8Y2hyDScdblUF6tFglAFbLUndko-eI

I'll defer to @steve-rivera on whether we want to give the instructor a list of resources either instead or in addition

steve-rivera:
Ah I didn't realize we already had instructor resources on course templates either. That's where I figured this would end up going. For now we can have instructor resources be an array and those can be shown on the course overview page like we've done for the Train the Trainer doc.

But I can definitely imagine a situation where we actually want to associate an instructor resource with a specific course activity or the course overview. So maybe instructor resources is an array of objects that contain the resource plus other attributes?

joncass:
@steve-rivera when you say "those can be shown on the course overview page like we've done for the Train the Trainer doc", how do you imagine that looking?

joncass:
Nice. A picture is worth a thousand words and all that.

steve-rivera:
If the array is very long that UI doesn't work quite as well and we may end up needing to add a resources tab at the top.

PR #58, opened 2020-06-23
Add IT_CERT_LINUX type to bundle spec

bpg130:
Similar to the `CLOUD_TERMINAL` resource, but there are no permissions that need to be set so there are no required attributes.


bpg130:
Per offline discussion, we will only be documenting this internally, so it does not make sense to publish it to the public spec.

PR #59, opened 2020-06-29
[IT Cert] Adding optional retake policies to CourseTemplate bundle spec

carinetorres:
## Quick Links
- [Buganizer Issue](https://b.corp.google.com/issues/159049160)


course-template-bundle-spec.md
```diff
@@ -182,3 +183,16 @@ activity_options | ✓        | array       | `activity_options` is an array of
...
+
+### Retake Policies
+
+A retake policy defines any required cooldown periods, retake limits, and retake windows applicable to a given `CourseTemplate` activity.
+
+attribute               | required | type        | notes
+----------------------- | -------- | ----------- | -----------------------------------------
+id                      | ✓        | string      | A unique identifier for this retake policy
+activity_type           | ✓        | enum        | The type of retakeable activity to which this retake policy applies; one of `Lab` or `Quiz`
+activity_subtype        |          | string      | Distinguishes a subset of a given type of retakeable activity to which this retake policy applies (i.e. `graded` for quiz)
```
joncass, line 195: nit:
```suggestion
activity_subtype        |          | string      | Distinguishes a subset of a given type of retakeable activity to which this retake policy applies (e.g. `graded` for quiz)
```

joncass: also we should indicate that these subtypes can be any string and must match the subtypes given on the activities

course-template-bundle-spec.md
```diff
@@ -182,3 +183,16 @@ activity_options | ✓        | array       | `activity_options` is an array of
...
+retake_cooldown         |          | array       | An array of `n` integers (greater than or equal to 0) specifying the required cooldown periods (in days) between consequent retakes; the last integer will be the cooldown period for all retakes after the `nth`
+retake_limit            |          | integer     | The total number of attempts (greater than 0) allowed for the given type and subtype of retakeable activity; required if `retake_window` below is specified
+retake_window           |          | integer     | An integer (greater than 0) specifying the period (in days) for which the `retake_limit` above applies
```
joncass, line 198: these are complicated enough that i think it would be nice to have a few examples

carinetorres: Done.

course-template-bundle-spec.md
carinetorres: Done.

course-template-bundle-spec.md
joncass: sorry, by examples i meant prose. like "for example, a retake policy of FOO means that a student can take a quiz up to N times per 24-hour period" or "for example, a retake policy of BAR means that after their first attempt, the student must wait X days before their second, and after their second attempt they must wait Y days before their third." ... stuff like that

carinetorres: Ah, done.

course-template-bundle-spec.md
```diff
@@ -179,6 +181,46 @@ attribute        | required | type        | notes
...
+retake_limit            |          | integer     | The total number of attempts (greater than 0) allowed for the given type and subtype of retakeable activity; required if `retake_window` below is  specified. For example, a retake_limit of `3` would indicate a student will *not* be allowed to retake the activity after a total of 3 attempts. If a student is within the retake_limit, they will only be allowed to retake the activity given they have waited for the appropriate cooldown period.
+retake_window           |          | integer     | An integer (greater than 0) specifying the period (in days) for which the `retake_limit` above applies. For example, a retake_limit of `3` paired with a retake_window of `1` would indicate a student can retake the activity only so long as no more than 3 attempts are made within a 1 day period.
```
joncass, line 226:
```suggestion
retake_window           |          | integer     | An integer (greater than 0) specifying the period (in days) for which the `retake_limit` above applies. For example, a retake_limit of `3` paired with a retake_window of `1` would indicate a student can retake the activity as long as no more than 3 attempts are made within a 1 day period.
```

course-template-bundle-spec.md
```diff
@@ -179,6 +181,46 @@ attribute        | required | type        | notes
...
+id                      | ✓        | string      | A unique identifier for this retake policy
+activity_type           | ✓        | enum        | The type of retakeable activity to which this retake policy applies; one of `Lab` or `Quiz`
+activity_subtype        |          | string      | Distinguishes a subset of a given type of retakeable activity to which this retake policy applies (e.g. `graded` for quiz); subtype must match one of the [allowed subtypes](#steps) given on the retakeable activity
+retake_cooldown         |          | array       | An array of `n` integers (greater than or equal to 0) specifying the required cooldown periods (in days) between consequent retakes; the last integer will be the cooldown period for all retakes after the `nth`; For example, a retake_cooldown of `[1, 2, 5]` would require a student is required to wait 1 day before their 1st retake, 2 days between their 1st and 2nd retakes, and 3 days between their 2nd and 3rd retakes, 3rd and 4th retakes, and so on.  If a student has waited for the appropriate coooldown period, they will only be allowed to retake the activity given they have not exceeded any retake limit.
```
joncass, line 224:
```suggestion
retake_cooldown         |          | array       | An array of `n` integers (greater than or equal to 0) specifying the required cooldown periods (in days) between consequent retakes; the last integer will be the cooldown period for all retakes after the `nth`; For example, a retake_cooldown of `[1, 2, 5]` would require a student to wait 1 day before their 1st retake, 2 days between their 1st and 2nd retakes, and 3 days between their 2nd and 3rd retakes, 3rd and 4th retakes, and so on.  If a student has waited for the appropriate coooldown period, they will only be allowed to retake the activity given they have not exceeded any retake limit.
```

carinetorres:
@jgan @eniskonuk @davetchen Forgot to get your review before merging, but please let me know if there are changes to make with retake policies.

course-template-bundle-spec.md
```diff
@@ -179,6 +181,46 @@ attribute        | required | type        | notes
...
+    activity_subtype: graded
+    retake_cooldown:
+      - 1
+    retake_limit: 3
+    retake_window: 1
+```
+
+attribute               | required | type        | notes
+----------------------- | -------- | ----------- | -----------------------------------------
+id                      | ✓        | string      | A unique identifier for this retake policy
```
jgan, line 221: If we're matching on `activity_type` and `activity_subtype`, we can only have one retake policy per combo, which makes this `id` field redundant... or at least simply decorative.

"Decorative" properties can absolutely be valuable in and of themselves if they act as an affordances for users; but the ContentBundle spec is targeted for machines, not humans. So I'd ask for a light justification about future use cases to keep the id around.

Alternatively, we could explicitly reference `retake_policy_id` on activities, and eschew the `subtype`.

course-template-bundle-spec.md
```diff
@@ -179,6 +181,46 @@ attribute        | required | type        | notes
 id               | ✓        | string      | A unique identifier for this step
 activity_options | ✓        | array       | `activity_options` is an array of dictionaries with the format:
 -- type          | ✓        | enum        | One of `lab`, `quiz`, `resource`
+-- subtype       |          | string      | A subordinate activity type; allowed `subtype` strings are contingent on the above `type`
```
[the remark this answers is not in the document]
joncass: i was imagining that the author could use any string for this and that it would not be used outside of retake context. i don't think retake_policy_id makes sense though since i would like them to use some kind of subtype-like descriptor.

course-template-bundle-spec.md
carinetorres: The presence of the additional ID is really to satisfy this check here: https://github.com/CloudVLab/git-whisperer/blob/master/whisperer/utils/yaml_merger.py#L134, less so because retaking requires it.

jgan: Ah, that's a good call in general; however...

1. Confusingly, it may be appropriate for "GitWhisperer format" to have an ID specifically for locale merging, but we don't need to include it as part of the Content Bundle spec. It's likely a GitWhisperer specific implementation detail. [Documentation for the GitWhisperer format is sorely needed as well.]

2. It doesn't look like any of these properties need to be localized, so we they don't need to be included in the locale specific yaml files (or merged).

joncass: yeah that's what i thought too, but carine was saying she saw the error, and b/160686692 makes me wonder if there's a bug in that as well...

course-template-bundle-spec.md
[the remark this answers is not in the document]
jgan: Agreed.

1. I'm cool with it being an arbitrary string instead of enum, let's talk replace the 'allowed `subtype` strings' comment to reflect that.

2. I'm hesitant to call it `subtype`. It's not really a "type", but a use context.

How do y'all feel about:

- `usage`
- `usage_type`
- `use_case`
- `style`?
- `kind`?
- `disposition`?
- ...?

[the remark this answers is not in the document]
joncass: sgtm. i don't know what the right word is. yours seem fine. here's more:
`context`?
`category`?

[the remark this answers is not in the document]
carinetorres: `category` resonates most with me... Will send out a PR with that change / corresponding description update.  I presume it makes sense to update `activity_subtype` below (in `Retake Policies`) to `activity_category` as well, so will also include that change.

course-template-bundle-spec.md
carinetorres: If that the behavior is intended to be (and will be fixed to be) such that non-localized objects in an array don't require an `id`, then SG, I will remove this `id` in the follow-up PR as well.

PR #61, opened 2020-07-08
Add method name to v2 lab

eniskonuk:
Based on discussion with @jgan & @akefeli and requests from the GCP Lab Architects, putting the `method_name` back into the spec.

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -405,8 +405,7 @@ title            | ✓        | locale dictionary          |
 maximum_score    | ✓        | integer                    | The maximum number of points this step can award.
 student_messages | ✓        | dictionary of locale dictionaries           | The keys are how the messages will be referenced in the code, and the values are locale dictionaries.
 services         | ✓        | array of resource services | An array of services that will be used in the code block. Each resource type specifies a set of allowed services.
-method_name      | ✓        | string                     | Name of the
-function/method in code block to be executed..
+method_name      | ✓        | string                     | Name of the function/method in code block to be executed.
 code             | ✓        | string                     | Code to be executed. See [below](#code) for more information.
 
 ```yml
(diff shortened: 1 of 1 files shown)


joncass:
I'm really curious what the use case for this is / how often people use it in v1. Can someone comment?

eniskonuk:
Most (>50%) of the V1 Labs in git use the GW feature that breaks out the `assessment:` into a separate yaml file and then having each of the code segments in their own `.rb` files in the `assessments` folder on git. That feature only works if each code segment has it's own `method_name` as GW figures out which `.rb` file is which code block through that attribute. So, it makes sense to continue that capability into V2.

joncass:
seems like what the author wants then is a way to specify what file to look for their code in, not the name of the method within the file. furthermore, i think that's an authoring/GW concern and not a bundle concern. so my proposal would be:
- do not add this to the bundle spec
- allow authors to specify code_filename which is a hint to GW but either doesn't make it into the final bundle or is ignored at unbundling time

akefeli:
> seems like what the author wants then is a way to specify what file to look for their code in, not the name of the method within the file. furthermore, i think that's an authoring/GW concern and not a bundle concern. so my proposal would be:
> 
> * do not add this to the bundle spec
> * allow authors to specify code_filename which is a hint to GW but either doesn't make it into the final bundle or is ignored at unbundling time

Jon main problem was keeping it backwards compatible, otherwise I agree. We want them to migrate to v2, they do not want to change activity tracking for every single lab.

eniskonuk:
Why? What problem not having `method_name` supported is solving? Doing all of the above suggestions is more work and I don't see the ROI.

joncass:
@akefeli - don't they already have to update the activity tracking code to accept the new function signature?

@eniskonuk - i don't want to have unnecessary required fields in our "API" (bundle spec) nor have to handle arbitrary method names in runtime. i agree it's a bit more work to do it right, but i think it will save work in the long term.

PR #62, opened 2020-07-16
pass percentage float?

akefeli:
do you guys care if this is float or int. its a float in database and we have been exporting it as float so far.

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -386,10 +386,10 @@ A lab has an Assessment, which in turn contains the Steps (checkpoints).
 
 #### Assessment
 
-attribute          | required | type    | notes
--------------------| -------- | --------| --------------------------------------
-passing_percentage | ✓        | float   | The percentage of total points the student must achieve to "pass" the lab.
-steps              | ✓        | array   | An array of [Steps](#steps)
+attribute          | required | type  | notes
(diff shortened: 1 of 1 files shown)


joncass:
float is fine with me. though then you should remove two of the dashes on line 390 :)

jgan:
Personally, I don't like it being a float.

With integers it's clear that the whole number represents a percentage.

As a float, it's a little ambiguous between whether 75% is `75` or `.75`. It also adds floating point arithmetic complications for (IMHO) unnecessary levels of precision.

My current opinion is: Let's change the export to cast to integers OR do a brief investigation to see if anyone is using sub-percent precision and change the underlying schema.

In either case, we may want to explicitly say the number must be between 0-100 in this documentation.

davetchen:
My `2.0` cents - leave it as a float to allow for greater precision (5 9's ?), but add documentation to make the range clear

eniskonuk:
I'm w/ @jgan on this one -- unnecessary level of precision that won't be
used and potentially confusing.

Thx
Enis Konuk
Google Cloud Learning
eniskonuk@google.com
+1 617-575-1643

akefeli:
I went with integer everywhere: json schema, db schema, code etc. no more headache from this hopefully

PR #63, opened 2020-07-17
[PE Fixit] Updating Lab v2 example

carinetorres:
Bundle added for testing in: https://github.com/CloudVLab/qwiklab-website/pull/5617.

## Quick Links
- [Buganizer Issue](https://b.corp.google.com/issues/153731813)


examples/lab-robust/v2/qwiklabs.yaml
```diff
@@ -1,32 +1,16 @@
...
-title:
-  locales:
-    en: Robust Lab Example
-    es: Ejemplo de Robust Lab
-
-description:
-  locales:
-    en: Seriously the best lab you've ever taken. Bar none.
-    es: En serio, el mejor lab que has tomado. Sin excepción.
+id: robust-bundle-v2
```
supersam654, line 4: Bundles shouldn't have an `id` field. We identify bundles with a unique ID but that ID is not part of the bundle itself. Please talk to Josh if you'd like more info on why.

carinetorres: Removed.

PR #64, opened 2020-07-29
Update lab-bundle-spec-v1.md

akefeli:
I'm actually leaning towards keeping max_duration as optional for v2 labs. I saw a lot of examples where author specified max_duration was different than how we would calculate it. It could be good to have it as an option

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -92,6 +92,8 @@ attribute              | required | type        | notes
 title                  | ✓        | locale dictionary |
 description            | ✓        | locale dictionary |
 duration               | ✓        | integer     | Amount of time it should take an average learner to complete the lab (in minutes)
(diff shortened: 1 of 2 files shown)


lab-bundle-spec.md
```diff
@@ -92,6 +92,8 @@ attribute              | required | type        | notes
 title                  | ✓        | locale dictionary |
 description            | ✓        | locale dictionary |
 duration               | ✓        | integer     | Amount of time it should take an average learner to complete the lab (in minutes)
+max_duration           |          | integer     | Maximum amount of time learner will be given to complete the lab (in minutes)
```
joncass, line 95: as discussed in chat, i think we should drop this

akefeli: done

PR #73, opened 2020-09-24
[Courses] Update CourseTemplate bundle spec with optional activity version

carinetorres:


joncass (commented):
@davetchen @jgan i thought we were planning on having the version be part of the content id when it was specified. i.e. you could either do library/slug or library/slug/version?

davetchen:
@carinetorres and I went back and forth on this. Having it as a separate field
- more clear for authors. they can already just specify "my-slug" then the library is inherited
- avoids possible parsing issues. library/slug is already just a "short form" for the actual Alexandria identifier 
- provides a consistent mechanism to specify version for a container entity if we decide to add that to the spec in the future

joncass:
I agree that we need to avoid ambiguity between library/slug and slug/version both having a single delimiter. I think it would be ok (and possibly even a good thing) to remove the slug-only "convenience", at least in the bundle spec itself. If we want git-whisperer to try to do something fancy to allow that (perhaps if you specify something with NO delimiters then it is slug, but if you do ONE delimiter then it is library/slug and TWO delimiters is library/slug/version??), then that's an option.

I'm not sure what you mean by "provides a consistent mechanism to specify version for a container entity if we decide to add that to the spec in the future" - can you rephrase?

jgan (approved):
TL;DR - I'm ok with this change from a practical usage perspective. But I do I have mixed feelings about further ingraining Alexandria implementation details into the content bundle spec.

----

A content bundle very purposefully does not include versioning information about *itself*. The content bundle only describes a learner experience, and it's the publishing system is the authority for how the bundles relate to each other; e.g. which of these bundles are "the same content", which bundle is the most recent version of "the same content", and potential more advanced questions like what "deprecated" content does this "new" content supersede.

When we're talking about Labs and Quizes, it's easy to keep publishing/versioning concerns out of the bundle. However, when we added container content entities that reference other content (e.g. Courses include Labs), the content bundle needs to be able to reference the publishing system's semantics for *other content*.

Currently we have two publishing systems, Alexandria and legacy Snapshots. Content bundles only reference Alexandria resources, and there are no plans for incorporating Snapshots as a lifeline for customers who don't fully migrate.

So, for now, I'm ok with `id` and `version` being tied directly the Alexandria's semantics; but overtime I would like to move towards using a more generic URI scheme for referencing published content.

Something along the lines of:

```yaml
# Type: ActivityReference
- type: lab
  ref: alexandria://the-library/the-slug

- type: lab
  ref: alexandria://the-library/the-slug/target-version

- type: lab
  ref: alexandria2://we-needed-to-overhaul/alexandria-id-schema/for-some-reason

- type: quiz
  ref: other_system://whatever-identifier-scheme-they-use
```

There's currently no other publishing system we're talking about integrating with, so I think this falls under the YAGNI principle.

However, as we start using Alexandria content in more of our systems (both Qwiklabs Platform and other Cloud Learning services), we'll start codifying patterns and shared code libraries for parsing and validating these references. At that point it would be easier to adopt this pattern.

jgan:
> I agree that we need to avoid ambiguity between library/slug and slug/version both having a single delimiter. I think it would be ok (and possibly even a good thing) to remove the slug-only "convenience", at least in the bundle spec itself.

💯 ... but we couldn't immediately simplify the Unbundler code path because of already published content bundles.

davetchen:
but wouldn't the more generic URI correspond to what Alexandria uses? e.g.
```
alexandria://libraries/gcp-enablement-content/content/TA-Data-Lake/versions/1.0
```

jgan:
> but wouldn't the more generic URI correspond to what Alexandria uses? e.g.

Yup! Absolutely, I was just being terse. I should have been more explicit to avoid confusion.

_Aside not directly relevant to this PR_: I happen to believe that shortened IDs without the constant segments (e.g. 'library', 'content') **_could_** be supported in Alexandria, because contents' increasingly specific hierarchy is a thoroughly embedded part of it's domain model; and any change to that would be along the lines of supporting a new `alexandria2://...` id format. BUT that would be a deliberate decision on Alexandria's side that would influence the list of supported reference schemes in the content bundle spec, not the other way around.

PR #74, opened 2020-09-30
add custom outputs for v2 labs

jenaepennie:
allow custom outputs for v2 labs and add it to the robust lab.

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -233,7 +233,7 @@ reference              | displayed as
 [PROJECT].default_zone | copyable text
 [PROJECT].console_url  | button
 
-Any custom outputs being generated with a `startup_script` can be reference as [PROJECT].<FILL-IN-OUTPUT-NAME>. The output will be displayed as a copyable text if provided within the student_visible_outputs.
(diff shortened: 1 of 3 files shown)


lab-bundle-spec.md
```diff
@@ -233,6 +233,8 @@ reference              | displayed as
 [PROJECT].default_zone | copyable text
 [PROJECT].console_url  | button
 
+Any custom outputs being generated with a `startup_script` can be reference as [PROJECT].<FILL-IN-OUTPUT-NAME>. The output will be displayed as a copyable text if provided within the student_visible_outputs.
```
jgan, line 236: I'd like this to add a segment for `startup_script` in the reference. That way there's no opportunity for reference collisions.

```
[PROJECT].startup_script.<MY-OUTPUT_NAME>

primary_project.startup_script.InstanceDns

# An author could do something very silly like name a custom output `default_zone`.
primary_project.startup_script.default_zone
```

This may seem contrived, but it's a good thing to avoid. Further more, we don't want to just validate that authors don't use "reserved references" at creation time, because we reserve the right to add new 1st class references in the future and don't want to break existing content.

supersam654 (approved):

PR #80, opened 2020-10-29
Add message back to assessment steps

jenaepennie:


lab-bundle-spec.md
```diff
@@ -598,4 +598,5 @@ The method `check` will be called with three keyword arguments:
 
 The method `check` should return a single hash with:
 - `:score`: the number of points the student earned.
+- `:message`: a message for only the lab creator to see when testing each assessment step within the Startup Lab Environment.
```
supersam654, line 601:
```suggestion
- `:message`: a message for only the lab creator to see when debugging each assessment step.
```

supersam654 (approved):

lab-bundle-spec.md
```diff
@@ -521,16 +521,16 @@ assessment:
           # Check for bucket
           found_bucket = ...
           unless found_bucket
-            return { score: 0, student_message: 'bucket_missing' }
+            return { score: 0, message: 'bucket_missing', student_message: 'bucket_missing' }
```
joncass, line 524: in the examples here and below we should not have the `message` match the `student_message` and look like a localization key

jenaepennie: What should it look like? `message` does have access to the localization key's message.

joncass: i am fine going either way. we should either have them be strings that make sense as log messages, or keys that point to log-specific sets of messages. either way we probably want to document in the description of the field how it works.

jenaepennie: [Updated Spec PR](https://github.com/CloudVLab/qwiklabs-content-bundle-spec/pull/81)

PR #82, opened 2020-11-13
Update G Suite resource description

supersam654:
* Rename G Suite to Google Workspace because branding changed and nobody is using this V2 resource yet
* Add outputs for the domain admin so they can be displayed like V1 labs


jgan (approved):
LGTM

When is the change scheduled in website?

supersam654:
Website changes are ongoing but the old resource never worked for V2 so I don't feel bad about merging this before we fully support it.

PR #86, opened 2021-01-06
bb terminal variants

robsideshow:
https://buganizer.corp.google.com/issues/174651983

Document the new terminal variants for the linux_terminal and windows_vm resources.  The actual diff is much smaller than it looks, since I shifted the position of the Windows VM section so all the BB resources are grouped together.

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -288,6 +288,11 @@ reference               | displayed as
 [DOMAIN].admin_password | copyable text
 
 
+
+
+
+#### BumperBoats Resources
+
 ##### Cloud Terminal (cloud_terminal)
 
 attribute   | required | type       | notes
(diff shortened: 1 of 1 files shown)


robsideshow:
cc @supersam654 @jgan

lab-bundle-spec.md
```diff
@@ -288,6 +288,11 @@ reference               | displayed as
 [DOMAIN].admin_password | copyable text
 
 
+
+
+
+#### BumperBoats Resources
```
bpg130, line 294: BumperBoats is an internal name and not something that I think we should publicize (similar to how Navy does not show up anywhere on this page).

I would either remove this header or rename it to something along the lines of `Virtual Machine Resources`.

lab-bundle-spec.md
```diff
@@ -317,13 +322,46 @@ startup_script.path |          | path  | Relative path to a directory tree with
     startup_script:
       path: startup.sh
 ```
+###### Variants for Linux Terminal
+
+The allowed `linux_terminal` variants are:
+- it_cert [default]  (access to internet via shared Cloud NAT IP address)
```
bpg130, line 328: Nit: Maybe clarify inbound vs outbound internet access?

`outbound access to internet via shared Cloud NAT IP address, no inbound internet access`.

bpg130 (changes requested):

lab-bundle-spec.md
robsideshow: Makes sense, though it sort of renews my vague ongoing confusion about the intended audience/purpose of this document.

lab-bundle-spec.md
robsideshow: how's that?

bpg130 (approved):

lab-bundle-spec.md
bpg130: LGTM

PR #89, opened 2021-02-05
Add the new looker_instance resource

supersam654:


lab-bundle-spec.md
```diff
@@ -331,6 +331,34 @@ The valid `reference`s for a `linux_terminal` resource are:
...
+permissions         | ✓        | array | Array of project/roles(array) pairs
+startup_script.path |          | path  | Relative path to a directory tree with the script contents.
+
+```yml
+  - type: looker_instance
+    id: looker
+    permissions:
+      - project: my_primary_project
+        roles:
+          - roles/editor
```
jgan, line 347: Sam, given our conversation, how do you feel about making the roles assigning implicit? We don't want to encourage the author to use the Looker instance to do anything other than the minimal BigQuery  (i.e. we don't want them treating it like a Cloud Shell).

```yaml
permissions:
  - project: my_primary_project
  - project: my_other_project

# ... or, more explicitly ...

permissions:
  - big_query_project: my_primary_project
  - big_query_project: my_other_project
```

Then in the documentation we say something like: "The Looker instance will have the BigQuery Admin permissions for any configured project."

supersam654: I thought we agreed to leave this as-is so it matches `cloud_terminal` and `gcp_user`.

If we're coming up with a new way to express these resources, I'd rather simplify the further and co with:

```
- type: looker_instance
  id: looker
  big_query_project: my_project
```

Note the lack of nesting or arrays

supersam654: After further discussion, we agreed that there are tradeoffs between the expressive yet verbose syntax in this PR compared with the more succinct options proposed in this comment thread. In the end, we decided to go with the verbose and expressive syntax because it matches the rest of the spec and allows for the greatest variety of labs in the future.

jgan (approved):

PR #91, opened 2021-02-24
Add IDE resource.

truongvi:


joncass:
We need to both be more confident in the exact definition of the resource and more along the path of production availability before we consider adding this to the spec, which is an external-facing document

supersam654 (approved):

PR #95, opened 2021-03-03
Add max_duration to bundle spec.

truongvi:
Allow creators to specify max_duration. 

https://buganizer.corp.google.com/issues/181710694

### DIFF
--- a/lab-bundle-spec.md
+++ b/lab-bundle-spec.md
@@ -50,6 +50,7 @@ description:
     en: "No, seriously. It's the best lab ever. You're going to love it!"
 
 duration: 60
+max_duration: 60
 level: intro
 tags: [sample, life-changing, gcp]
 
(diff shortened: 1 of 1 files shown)


lab-bundle-spec.md
```diff
@@ -50,6 +50,7 @@ description:
     en: "No, seriously. It's the best lab ever. You're going to love it!"
 
 duration: 60
+max_duration: 60
```
joncass, line 53: i know it's just an example, but would be good to use different values so that creators don't think they're supposed to be the same thing
```suggestion
max_duration: 90
```

lab-bundle-spec.md
```diff
@@ -94,6 +95,7 @@ attribute              | required | type        | notes
 title                  | ✓        | locale dictionary |
 description            | ✓        | locale dictionary |
 duration               | ✓        | integer     | Amount of time it should take an average learner to complete the lab (in minutes)
+max_duration           |          | integer     | Maximum duration time of a lab
```
davetchen, line 98:
```suggestion
max_duration           |          | integer     | Maximum duration time of a lab, displayed in student timer
```

akefeli: can you also add "(in minutes)" to end

akefeli (approved):

lab-bundle-spec.md
truongvi: Done.

PR #99, opened 2021-05-26
Update lab bundle spec for new workspace options

stellamazeika:
Working on the documentation side of the spec - starting with these two and will roll out more as needed.


supersam654 (approved):
LGTM. While you're here, you may want to add the other URLs just because updating the spec has a bunch of steps and you probably only want to go through them once.

PR #100, opened 2021-06-01
Specify when we pre-fill email addresses

supersam654:


github-lab-bundle-spec.md
```diff
@@ -334,6 +336,8 @@ reference              | displayed as
  [USER].gmail_url      | button
  [USER].drive_url      | button
  [USER].calendar_url   | button
+  
+All of the URLs will take the student to a new page where they can sign in with their temporary credentials and then get redirected to the requested Workspace service. When signing in, the temporary email address will be pre-filled but the student will still need to enter a password. If there are multiple temporary students in a lab (rare), we recommend creating different buttons for each one so each sign-in experience has the correct pre-filled email address.
```
carinetorres, line 340: As a non-author I'm confused (idk if anyone else would be), but by "creating different buttons" do you mean adding multiple for example `[USER].gmail_url` references? Perhaps consider adding an example snippet of the yaml with such?

supersam654: Thanks for the feedback! I added an example below this section. Let me know if that makes more sense or if it's still confusing.

carinetorres (approved):

github-lab-bundle-spec.md
carinetorres: LGTM thanks for updating!

PR #106, opened 2021-06-14
Update github-lab-bundle-spec.md

akefeli:


github-lab-bundle-spec.md
```diff
@@ -85,9 +85,6 @@ duration: 60
 level: intro
 tags: [sample, life-changing, gcp]
 
-# The primary instruction content for this lab
```
hanysf, line 88: So, this is not configurable anymore. We just require creators to place their instructions under a folder called `instructions`? If so, I am curious as to what caused this change.

github-lab-bundle-spec.md
```diff
@@ -142,16 +139,13 @@ legacy_display_options |          | array       | Elements to hide/show in ql-la
...
-type      | ✓        | enum              | [See list of valid types below]
-uri       | ✓        | string            | Paths to files in the bundle
+Place your instruction files under "instructions" folder. Name of the file MUST be same as the locale key.
 
-```yaml
-instruction:
-  type: html
-  uri: instructions/en.html
 ```
+my-lab/instructions/en.md
```
eniskonuk, line 145: Don't you need the `-instruction:` key before the files?

hanysf: Seems like the configuration section in qwiklabs.yaml is gone. This is an example of how the different locale files look like under the `instructions` folder.

github-lab-bundle-spec.md
akefeli: it is not a change, it's always been like this. GW always rewrites instruction block itself regardless of what you put there. Writing yourself is noop at best, or sometimes confusing because removing an instruction there won't actually remove it

github-lab-bundle-spec.md
akefeli: I didn't mean that as part of yaml, I'll clarify it

hanysf (approved):

PR #108, opened 2021-07-01
Bundle updates for qwiklabs scripts

bpg130:
Mark the script type as required, which in the code, it has been all along and document the new option for the `qwiklabs` script type.

### DIFF
--- a/github-lab-bundle-spec.md
+++ b/github-lab-bundle-spec.md
@@ -235,7 +235,7 @@ environment:
 
 attribute                        | required | type    | notes
 ---------------------------------| -------- | ------- | --------------------------------------
-startup_script.type              |          | string  | The type of startup script. Only `deployment_manager` is supported.
+startup_script.type              | ✓        | string  | The type of startup script. Only `deployment_manager` and `qwiklabs` are supported.
 startup_script.path              |          | path    | Relative path to a directory tree with the script contents.
 startup_script.custom_properties |          | array   | Array of pairs. See below for details.
 ssh_key_user                     |          | string  | If this project should use a user's SSH key, the id of that user.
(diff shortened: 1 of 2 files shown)


bpg130:
cc: @supersam654

robsideshow:
I believe the current desired procedure is that humans/authors take their cues on what to commit to github from `main/github-lab-bundle-spec.md`, while the the file `api/lab-bundle-spec.md` is meant to define the *output* of GW (aka *input* of Website), and we never want an author to lay eyes upon it. :)  Of course, the two are almost identical. :) TLDR, I _think_ you also need to update https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/main/github-lab-bundle-spec.md#gcp-project-gcp_project
correct, @eniskonuk ?

bpg130:
> I believe the current desired procedure is that humans/authors take their cues on what to commit to github from `main/github-lab-bundle-spec.md`, while the the file `api/lab-bundle-spec.md` is meant to define the _output_ of GW (aka _input_ of Website), and we never want an author to lay eyes upon it. :) Of course, the two are almost identical. :) TLDR, I _think_ you also need to update https://github.com/CloudVLab/qwiklabs-content-bundle-spec/blob/main/github-lab-bundle-spec.md#gcp-project-gcp_project
> correct, @eniskonuk ?

Done.

github-lab-bundle-spec.md
```diff
@@ -235,7 +235,7 @@ environment:
 
 attribute                        | required | type    | notes
 ---------------------------------| -------- | ------- | --------------------------------------
-startup_script.type              |          | string  | The type of startup script. Only `deployment_manager` is supported.
```
robsideshow, line 238: What do you think of adding a link to go/ql-scripts? Up to you.

robsideshow (approved):
lgtm

github-lab-bundle-spec.md
bpg130: I am a little hesitant because this is a public repo and the go link is only valid within Google. I think it may be better to someday move the content of go/ql-scripts to this repo.

PR #117, opened 2021-09-01
add documentation about escaping HTML within code blocks

shoffing:


html/html-spec.md
```diff
@@ -129,6 +129,30 @@ Would be rendered as:
   }
 ```
 
+### Escaping Content
```
carinetorres, line 132: While you're editing this page, can you modify above section about `language` to note the default will be plaintext?

html/html-spec.md
```diff
@@ -129,6 +129,30 @@ Would be rendered as:
...
+
+```html
+<ql-code-block language="html">
+&lt;h1&gt;What is a monad?&lt;/h1&gt;
+&lt;p&gt;A monad is just a monoid in the category of endofunctors.&lt;/p&gt;
+&lt;img src="monad.png" alt="Monad diagram"&gt;
+</ql-code-block>
+```
+
+will render as:
```
carinetorres, line 148: Thoughts on using a screenshot of the actual code block users see instead?  Might be more helpful for authors to actually see what that looks like.

html/html-spec.md
shoffing: I thought `qwiklabs-content-bundle-spec` is "git-whisperer agnostic", so that doesn't apply here - `<ql-code-block>` will actually automatically detect the language if it's not specified.

html/html-spec.md
shoffing: Sounds good! Done

html/html-spec.md
shoffing: Actually maybe not, I see in ql-variable i included some documentation on the markdown syntax. I should do something similar here for ql-code-block.

carinetorres (approved):

html/html-spec.md
shoffing: Done, added a section about Markdown. Please review

carinetorres (approved):

PR #152, opened 2023-04-20
Add hinting

stellamazeika:
Add hinting to the lab spec

### DIFF
--- a/html/html-spec.md
+++ b/html/html-spec.md
@@ -339,8 +339,30 @@ Result (lab started, same as the previous example):
 
 ![example preview](./images/ql-variable-03.png)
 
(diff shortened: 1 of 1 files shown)


html/html-spec.md
```diff
@@ -339,8 +249,29 @@ Result (lab started, same as the previous example):
 
 ![example preview](./images/ql-variable-03.png)
 
-### `<ql-activity-tracking>`
+## `<ql-hint>`
```
shoffing, line 252: I think these changes should happen in google3, then Copybara will copy it over to here. https://source.corp.google.com/piper///depot/google3/cloud/training/qwiklabs/content_bundle/g3doc/html/html-spec.md

